PR #4, opened 2024-12-09
[feature] Solarwinds Extension

david-sevcik:
### Description

Adds new solarwinds extension which is sending heartbeat at specific interval (30s) to signal liveliness in SWO. The heartbeat signal is an uptime metric `sw.otelcol.uptime`.
It's decorated with a collector name, which is configurable in the config.
It also take over endpoint configuration from the solarwinds exporter and provides this configuration to the exporter at runtime. Hence the exporter is now dependent on the extension and doesn't start without it.

### Tech Notes

- The new extension was added to the built distribution - should be available for evrybody which will use the Solarwinds OTEL Collector
- The GO version was raised because if some new features used in the codebase
- The configuration, readme and tests for exporter were updated to not check the removed configuration
   - Some genereted tested had to be skipped, because the exporter now requires the extension to run, but the generated tests cannot be configured to provide the extension to the tested exporter
- There can be more solarwinds extension instances - if that's the case, the exporter needs the instance ID of specific extension to connect, otherwise it fails - it needs to know which extension instance to ask for the endpoint configuration
- there are some small fixes of pipeline and in the code


exporter/solarwindsexporter/solarwinds_exporter.go
```diff
@@ -34,111 +36,166 @@ const (
...
+			}
+
+			// Otherwise, store it to the slice.
+			foundExtensions = append(foundExtensions, swiExtension)
+		}
+	}
+
+	// If no extension name configured and there is only one
+	// found matching the type, return it.
+	if len(foundExtensions) == 1 && cfgExtensionID != nil {
```
simek-m, line 142:
```suggestion
	if len(foundExtensions) == 1 && cfgExtensionID == nil {
```

adschr:
I believe these changes are incoming anyway, being in draft at this point, but to be sure: can we please adjust the PR name and add a proper description on what is being added/what the changes are/hints to reviewers on what to give explicit feedback on?

PR template incoming soon to standardize all of this :).

cmd/solarwinds-otel-collector/go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/solarwinds/solarwinds-otel-collector
 
-go 1.22.7
-
-toolchain go1.22.9
+go 1.23.4
```
adschr, line 3: is this required due to transitive imports from the extension? So this can't run on lower version?

exporter/solarwindsexporter/factory.go
```diff
@@ -45,7 +45,10 @@ func createMetricsExporter(
 		return nil, fmt.Errorf("unexpected config type: %T", cfg)
 	}
 
-	metricsExporter := newExporter(ctx, exporterCfg, settings, metricsExporterType)
+	metricsExporter, err := newExporter(exporterCfg, settings, metricsExporterType)
+	if err != nil {
+		return nil, err // TODO: Wrap.
```
adschr, line 50: just adding explicit comment to make sure the TODO is either resolved or comment removed

exporter/solarwindsexporter/config.go
```diff
@@ -68,12 +42,45 @@ type Config struct {
 	QueueSettings exporterhelper.QueueConfig `mapstructure:"sending_queue"`
 	// Timeout configures timeout in the underlying OTLP exporter.
 	Timeout exporterhelper.TimeoutConfig `mapstructure:"timeout,squash"`
```
adschr, line 44: these OTLP settings are kept in the exporter and not added/reused within the extension as it is presumed those won't be needed for the collector liveness metrics?

exporter/solarwindsexporter/config.go
```diff
@@ -68,12 +42,45 @@ type Config struct {
...
-// NewDefaultConfig creates a new default configuration.
+// ExtensionAsComponent tries to parse `extension` value of the form 'type/name'
+// or 'type' from the configuration to [component.ID].
+// It fails with an error if it doesn't follow this form or the 'type' part
+// is not a valid [component.Type].
 //
-// Warning: it doesn't define mandatory `Token` and `DataCenter`
-// fields that need to be explicitly provided.
+// Safety: it PANICS if `extension` is empty.
+func (cfg *Config) ExtensionAsComponent() (component.ID, error) {
```
adschr, line 57: do we need this function public?

simek-m: @david-sevcik Great point, yes, it's better to use that one (it seems, they handle same additional) edge cases. I knew they must have it implemented but missed this one can be used by us.

@adschr I think it can be public but let's limit the API surface.

exporter/solarwindsexporter/solarwinds_exporter.go
```diff
@@ -37,108 +40,176 @@ const (
...
+
+	// Only allow the type of the [solarwindsextension].
+	if extensionID != nil &&
+		extensionID.Type() != solarwindsextension.NewFactory().Type() {
+		return fmt.Errorf("unexpected extension type: %s", extensionID.Type())
+	}
+
+	swiExtension := findExtension(host.GetExtensions(), extensionID)
+	if swiExtension == nil {
+		return errors.New("solarwinds extension not found")
```
adschr, line 92: nit: maybe mention the extension ID for better debug experience?

exporter/solarwindsexporter/solarwinds_exporter.go
```diff
@@ -37,108 +40,176 @@ const (
...
+	endpointCfg := swiExtension.GetEndpointConfig()
+
+	// Get token from the extensions.
+	token := endpointCfg.Token()
+	swiExporter.config.ingestionToken = token
+
+	// Get URl from the extension.
+	url, err := endpointCfg.Url()
+	if err != nil {
+		return fmt.Errorf(": %w", err)
```
adschr, line 104: I suppose some message might get here :)

exporter/solarwindsexporter/config.go
simek-m: Changed in https://github.com/solarwinds/solarwinds-otel-collector/pull/4/commits/4b11ae00436867cf45c6ca5f9bd257fb89fdbb15 (the behaviour seems to be identical and it simplified empty value handling too).

extension/solarwindsextension/internal/heartbeat.go
```diff
@@ -0,0 +1,138 @@
...
+			return err
+		}
+	}
+
+	return h.exporter.push(ctx, md)
+}
+
+func (h *Heartbeat) decorateResourceAttributes(resource pcommon.Resource) error {
+	if h.collectorName != "" {
+		resource.Attributes().PutStr("collector_name", h.collectorName)
```
adschr, line 135: Looking at the [Resource](https://opentelemetry.io/docs/specs/semconv/resource/)/[Attributes](https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/) naming conventions, we might have more specific round of discussion about the newly added functionality naming. For starters, it might be recommended to go with namespace approach (`collector.name`, ...) but there is more to it, metric naming included.

adschr: For the record, as discussed offline, let's name:
- metric as `sw.otelcol.uptime`
- res. attribute as `sw.otelcol.collector.name`
To avoid any potential future clashing with pretty general `collector` namespace, since these changes are bound to sw context.

extension/solarwindsextension/internal/heartbeat.go
```diff
@@ -0,0 +1,138 @@
...
+
+	cancel           context.CancelFunc
+	startShutdownMtx sync.Mutex
+
+	metric        *UptimeMetric
+	exporter      *Exporter
+	collectorName string
+}
+
+var alreadyRunningError = errors.New("heartbeat already running")
```
simek-m, line 40: Minor: I'd follow the naming convention for sentinel errors from the [stdlib](https://pkg.go.dev/os/exec#pkg-variables), i.e. `ErrAlreadyRunning` and make it public as the error is returned from a public function.

extension/solarwindsextension/README.md
```diff
@@ -0,0 +1,41 @@
...
+| Status        |           |
+| ------------- |-----------|
+| Stability     | [development]  |
+| Distributions | [] |
+| Issues        | [![Open issues](https://img.shields.io/github/issues-search/solarwinds/solarwinds-otel-collector?query=is%3Aissue%20is%3Aopen%20label%3Aextension%2Fsolarwinds%20&label=open&color=orange&logo=opentelemetry)](https://github.com/solarwinds/solarwinds-otel-collector/issues?q=is%3Aopen+is%3Aissue+label%3Aextension%2Fsolarwinds) [![Closed issues](https://img.shields.io/github/issues-search/solarwinds/solarwinds-otel-collector?query=is%3Aissue%20is%3Aclosed%20label%3Aextension%2Fsolarwinds%20&label=closed&color=blue&logo=opentelemetry)](https://github.com/solarwinds/solarwinds-otel-collector/issues?q=is%3Aclosed+is%3Aissue+label%3Aextension%2Fsolarwinds) |
+
+[development]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#development
+<!-- end autogenerated section -->
+
+SolarWinds Extension is required sidecar for the [Solarwinds Exporter](../../exporter/solarwindsexporter).
```
simek-m, line 13: I'd be cautious about the term "sidecar" here - it could be confusing to our users with a K8s background where it means a specific thing.

extension/solarwindsextension/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+It provides these features:
+
+- Hearbeat signal
```
simek-m, line 17:
```suggestion
- Heartbeat signal
```
Typo (hate that word).

extension/solarwindsextension/internal/heartbeat.go
simek-m: Yes, done in https://github.com/solarwinds/solarwinds-otel-collector/pull/4/commits/413b882ddadc700ba7604ad8cbd28523c0fc742a

extension/solarwindsextension/internal/heartbeat.go
simek-m: Renamed in 5c987888becb7af855d10c766bd913e6cac1ecd1

exporter/solarwindsexporter/README.md
```diff
@@ -13,25 +13,24 @@ SolarWinds Exporter is a convenience wrapper around [OTLP gRPC Exporter](https:/
 
 ## Getting Started
 
-You just need to include the SolarWinds Exporter in your exporter definitions and provide the following minimal configuration:
+You just need to include the SolarWinds Exporter in your exporter definitions and no additional configuration is needed. It always needs to be used together with the [Solarwinds Extension](../../extension/solarwindsextension).
 
 ```yaml
 exporters:
+  solarwinds: {}
```
adschr, line 20: nit: not sure how to interpret the `{}` from reader perspective

exporter/solarwindsexporter/config.go
```diff
@@ -90,18 +83,6 @@ func NewDefaultConfig() component.Config {
 
 // Validate checks the configuration for its validity.
 func (cfg *Config) Validate() error {
-	if cfg.DataCenter == "" && cfg.EndpointURLOverride == "" {
```
adschr, line 93: nit: would it maybe make sense to perform some extension sanity check? just to clear out straight wrong input from the top

extension/solarwindsextension/README.md
```diff
@@ -0,0 +1,41 @@
...
+```yaml
+extensions:
+  solarwinds:
+    token: "YOUR-INGESTION-TOKEN"
+    data_center: "na-01"
+    collector_name: "Collector Display Name"
+```
+- `token` (mandatory) - You can generate your token in your SWO (SolarWinds Observability) account under _Settings / API Tokens / Create API Token_. The type is "Ingestion". You can find the complete documentation [here](https://documentation.solarwinds.com/en/success_center/observability/content/settings/api-tokens.htm).
+- `data_center` (mandatory) - Data center is the region you picked during the sign-up process. You can easily see in URLs after logging in to SolarWinds Observability SaaS - it's either `na-01`, `na-02` or `eu-01`. Please refer to the [documentation](https://documentation.solarwinds.com/en/success_center/observability/content/system_requirements/endpoints.htm#Find) for details.
+- `collector_name` (mandatory) - this is the name under which the collector will be shown in the SWO UI. Doesn't have to be unique.
```
adschr, line 37: I'd skip the part about collector being shown in SWO. At this point, the name is solely for differentiation or filtering of data

extension/solarwindsextension/README.md
```diff
@@ -0,0 +1,41 @@
...
+<!-- end autogenerated section -->
+
+SolarWinds Extension is required sidecar for the [Solarwinds Exporter](../../exporter/solarwindsexporter).
+
+It provides these features:
+
+- Hearbeat signal
+  - It's a standard metric: `sw.otecol.uptime`
+    - The value is a time from the start of the collector in seconds
+  - This signal is necessary for SWO to detect the collector when installed and to determine if it's still alive.
```
adschr, line 20: SWO -> SolarWinds Observability SaaS (please update in all places)

extension/solarwindsextension/internal/config.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// Validate checks the configuration for its validity.
+func (cfg *Config) Validate() error {
+	if cfg.DataCenter == "" && cfg.EndpointURLOverride == "" {
+		return ErrMissingDataCenter
+	}
+
+	if cfg.Insecure && cfg.EndpointURLOverride == "" {
+		return ErrInsecureInProd
```
adschr, line 71: is this covered by tests? (please disregard if we drop the insecure config after all)

adschr (dismissed):
left some nits, in general LGTM. Let's proceed only after we validate the need for the `insecure` setting.

exporter/solarwindsexporter/config.go
simek-m: @adschr Do you think validate if the `DataCenter` value is a valid DC ID?

extension/solarwindsextension/internal/config.go
simek-m: It is now: https://github.com/solarwinds/solarwinds-otel-collector/pull/4/commits/53e7753aaf87cd430d131c5363ba430d54c7ad46 (just a single, simple testcase).

exporter/solarwindsexporter/config.go
adschr: I meant validity of `extension` field (as the rest was removed)

simek-m: Done in https://github.com/solarwinds/solarwinds-otel-collector/pull/4/commits/1f6e7ea828e565b4754f7e514d410439e1ef59c7

exporter/solarwindsexporter/README.md
adschr: to me yes, as I tend to more YAML-puritan notation, but if this is somehow standardized 'tis a gray zone and I'll defer to your choice 👍

jirkajanecek (approved):

jirkajanecek (approved):

adschr (approved):

PR #6, opened 2024-12-12
[feature] Docker publishing and GH releases

adschr:
This PR adds workflow for docker publishing. Inspired by [swi-k8s-collector](https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/.github/workflows/buildAndDeploy.yml), but with some exceptions:
* omitting windows builds for now
* replacement of deprecated "::set-output" macros to the GIT_OUTPUT publishing (this actually already emitted some warnings to the pipeline). More reading: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Workflow is sensitive to pushing to `refs/tags/` github ref, with specific 'v' prefix. The `${GITHUB_REF#refs/tags/v*}` operation extracts the numeric value of the tag for use as docker tags (the GH tags should be in format 'vX.Y.Z', while the Docker ones just X.Y.Z.)

Workflow is chained and goes as following:
* deploy and publish to docker hub linux amd/arm
* create and publish docker manifests for those
* create GH release with generated notes from the commits in between the latest and current tag

The GitHub tag value has not been extracted to separate action yet as the oneliner is not that complex. But I defer that to feedback.

**NOTE:** Readme did not require update, the note on publishing and releases was already there


simek-m (approved):
It LGTM:)

I was just thinking it might be possible to do without qemu and manifests with `buildx` - something like [this random gist does](https://gist.github.com/AverageMarcus/78fbcf45e72e09d9d5e75924f0db4573) thanks to Go's easy cross-compilation (and CGO_ENABLED=0) but if my understanding is correct, the `RUN` steps would still require some kind of emulation / runner.

akumetsu183 (approved):

PR #8, opened 2024-12-18
[test] Integration tests for collector

jirkajanecek:
#### Description
- Introducing integration tests
- Tests are docker based using test containers for Go with following scheme
    - [[telemetry generator](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/cmd/telemetrygen) container] => [emitting Solarwinds collector aka `sut`] => [receiving SolarWinds collector]
- Following signals are tested on delivery point:
    - metrics
    - traces
    - logs
    - heartbeat metrics (aka `sw.otelcol.uptime`) as product of Solarwinds extension 
- integration tests are part of GHA workflow
  - utilized in `build_and_deploy` job to leverage the potential to reuse the image w/o any time overhead. Having it in separate job would require image exporting as an artifact and then downloading it in the next one.
  - added latest tag to the build and fixed the build-time tag to contain also the image name, not only the version tag itself

**UPDATE:** In process it was discussed that the test level from the point of view of the collector itself are more of e2e in nature - naming within changes has been updated to reflect this.


internal/e2e/go.mod
```diff
@@ -0,0 +1,69 @@
+module github.com/solarwinds/solarwinds-otel-collector/internal/e2e
+
+go 1.23.4
+
+require (
+	github.com/stretchr/testify v1.10.0
+	github.com/testcontainers/testcontainers-go v0.34.0
+	go.opentelemetry.io/collector/pdata v1.21.0
```
simek-m, line 8: The version of `go.opentelemetry.io/collector/pdata` is not in sync with other modules. It should be `v1.19.0`.

internal/e2e/instrumentation.go
```diff
@@ -0,0 +1,131 @@
+// Copyright 2024 SolarWinds Worldwide, LLC. All rights reserved.
```
simek-m, line 1: Minor / subjective: `infrastructure.go` sounds a bit too vague to me - maybe `containers.go` or `collector_runner.go`, ...

internal/e2e/integration_test.go
```diff
@@ -0,0 +1,292 @@
+// Copyright 2024 SolarWinds Worldwide, LLC. All rights reserved.
```
simek-m, line 1: `internal/e2e/integration_test.go`: Is it `integration` or `e2e`? I think it can be both.

internal/e2e/README.md
```diff
@@ -0,0 +1,15 @@
+# Integration Tests
+The purpose of the integration tests is to validate the overall integrity of SolarWinds OpenTelemetry Collector distribution and the components used within, verifying correct handling of all supported signals.
+
+## How to Develop
+For local test runs it is required to have local build of SolarWinds OpenTelemetry Collector tagged as `solarwinds-otel-collector:latest`.
+
+Tests then can be run with `go test -tags=integration` command.
```
simek-m, line 7: It'd be nice to have this in the `Makefile`.

internal/e2e/go.mod
adschr: indeed, updated

internal/e2e/integration_test.go
adschr: yop, from the collector's perspective it can be e2e, yet the surrounding `integration` term usage around is predominant indeed. I think with naming we might even go after what it acually tests - sort of `signals_processing_test.go` or similar as that's what we are interested in in the end. 

Still, we should make the final call and potentially tidy up in the build tags, job names etc. so there is clear line drawn of what this level is (and is supposed to be contained within the package).

simek-m (approved):

internal/e2e/signals_processing_test.go
```diff
@@ -0,0 +1,292 @@
...
+	require.Equal(t, expectedCount, lgs.LogRecordCount())
+}
+
+func evaluateHeartbeatMetric(
+	t *testing.T,
+	ms pmetric.Metrics,
+) {
+	require.GreaterOrEqual(t, ms.ResourceMetrics().Len(), 1, "there must be at least one metric")
+	atts := ms.ResourceMetrics().At(0).Resource().Attributes()
+	v, available := atts.Get("sw.otelcol.collector.name")
```
simek-m, line 260: It could possibly utilise https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/pdatatest but I had issues with dependencies when I tried to use that.

internal/e2e/testdata/emitting_collector.yaml
```diff
@@ -0,0 +1,43 @@
...
+      exporters: [solarwinds]
+    traces:
+      receivers: [otlp]
+      exporters: [solarwinds]
+    logs:
+      receivers: [otlp]
+      exporters: [solarwinds]
+
+receivers:
+  # Just as creative source.
```
simek-m, line 15: Not sure what `creative` means here, "emitting", "producing"?

simek-m (dismissed):

internal/e2e/signals_processing_test.go
adschr: looks like dependencies should not be an issue in this separate package, yet I am a bit suspicious of flaky results, since the metrics count is dependent on the actual runtime (here this is mitigated by the check for at least one metric).. 
nevertheless package mentioned usage is a good idea, I'll work it into a follow-up.

simek-m: @adschr I meant something different - there were some conflicts with OTEL dependencies when I tried to use the `pdatatest` module for `heartbeat_test.go` and I couldn't compile it with that. I'm getting a different error now, though.

internal/e2e/signals_processing_test.go
```diff
@@ -0,0 +1,292 @@
...
+	resourceAttributeName  = "resource.attributes.testing_attribute"
+	resourceAttributeValue = "testing_value"
+)
+
+func TestMetricStream(t *testing.T) {
+	ctx := context.Background()
+
+	net, err := network.New(ctx)
+	require.NoError(t, err)
+	testcontainers.CleanupNetwork(t, net)
```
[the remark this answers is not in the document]
jirkajanecek: This is written as doc in function call
```
// CleanupNetwork is a helper function that schedules the network to be
// removed when the test ends.
// This should be the first call after NewNetwork(...) in a test before
// any error check. If network is nil, its a no-op.
```

Simply said, it should be removing resources when test is over.

I think, this API is "only proxy" to Docker calls, so when you register for clean up it will remove it from running GHA image - at least I think it is like that.

internal/e2e/testdata/receiving_collector.yaml
```diff
@@ -0,0 +1,25 @@
...
+    protocols:
+      grpc: 
+        endpoint: :17016
+exporters:
+  debug:
+    verbosity: detailed
+    sampling_initial: 5
+    sampling_thereafter: 200
+  file:
+    path: /tmp/result.json
```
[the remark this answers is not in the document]
jirkajanecek: I think it is. 

Tests only test one signal (plus heartbeat metric) in a time. It is given by their nature. In addition, Telemetry generator is configured to send only one signal in a time.

I could probably configure non metric signals output to different file (by using aliasing) to separate it from heartbeat metric, but I haven't. Output content is separated in test.

internal/e2e/signals_processing_test.go
```diff
@@ -0,0 +1,292 @@
...
+	require.NoError(t, err)
+	testcontainers.CleanupContainer(t, rContainer)
+
+	eContainer, err := runTestedSolarWindsOTELCollector(ctx, net.Name)
+	require.NoError(t, err)
+	testcontainers.CleanupContainer(t, eContainer)
+
+	cmd := []string{
+		"metrics",
+		"--metrics", strconv.Itoa(samplesCount),
```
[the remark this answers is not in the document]
jirkajanecek: TLDR: I think it is, in a way how evaluation is done.

Metrics are generated in TelemetryGen provided by contrib.

It is deterministic in a way how it is evaluated. Gen is configured to send N samples and that count is tested - aka all arrive. I don't test exact values delivered.

Only explicit evaluation is done on resource attributes, which are explicitly configured to specific value.

simek-m (approved):

internal/e2e/signals_processing_test.go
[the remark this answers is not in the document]
simek-m: @david-sevcik I thought the same but it is actually deferred inside (the registers a cleanup function after the test).

PR #17, opened 2025-01-03
feature: optional custom attributes for extension

akumetsu183:
#### Description
solarwindsextension now has optional attributes

If any are specified, the uptime metric will be enriched with the given attributes, and their values. 

#### Testing
Tested on the docker image.

### DIFF
--- a/extension/solarwindsextension/internal/config.go
+++ b/extension/solarwindsextension/internal/config.go
@@ -39,7 +39,7 @@ type Config struct {
 	// ⚠️ Warning: For testing purpose only.
 	// EndpointURLOverride sets OTLP endpoint directly, it overrides the DataCenter configuration.
 	EndpointURLOverride string            `mapstructure:"endpoint_url_override"`
-	Attributes          map[string]string `mapstructure:"attributes"`
+	Resource            map[string]string `mapstructure:"resource"`
 }
 
 var (
--- a/extension/solarwindsextension/internal/config.go
+++ b/extension/solarwindsextension/internal/config.go
@@ -38,7 +38,8 @@ type Config struct {
 	CollectorName string `mapstructure:"collector_name"`
 	// ⚠️ Warning: For testing purpose only.
 	// EndpointURLOverride sets OTLP endpoint directly, it overrides the DataCenter configuration.
-	EndpointURLOverride string `mapstructure:"endpoint_url_override"`
+	EndpointURLOverride string            `mapstructure:"endpoint_url_override"`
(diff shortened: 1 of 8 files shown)


extension/solarwindsextension/internal/config.go
```diff
@@ -38,7 +38,8 @@ type Config struct {
 	CollectorName string `mapstructure:"collector_name"`
 	// ⚠️ Warning: For testing purpose only.
 	// EndpointURLOverride sets OTLP endpoint directly, it overrides the DataCenter configuration.
-	EndpointURLOverride string `mapstructure:"endpoint_url_override"`
+	EndpointURLOverride string            `mapstructure:"endpoint_url_override"`
+	Attributes          map[string]string `mapstructure:"attributes"`
```
akumetsu183, line 42: The naming can be changed, but this made sense to me so I went with it.

extension/solarwindsextension/testdata/full.yaml
```diff
@@ -2,3 +2,6 @@ token: "TOKEN"
 data_center: "na-01"
 collector_name: "test-collector"
 endpoint_url_override: "127.0.0.1:1234"
+attributes:
+  att1: "custom_attribute_value_1"
+  att2: "custom_attribute_value_2"
```
simek-m, line 7: Minor: missing new line.

jirkajanecek (approved):
Do we want to extend e2e tests with this new feature of sending custom attributes?

extension/solarwindsextension/internal/config.go
akumetsu183: Renamed to Resource

extension/solarwindsextension/testdata/full.yaml
akumetsu183: Fixed

jirkajanecek (approved):
I think it is good.

adschr:
> Do we want to extend e2e tests with this new feature of sending custom attributes?

Loos like good idea to me as I do see test coverage only the config extraction, not the utilization - @akumetsu183 could you please extend the e2e test? Or if it is possible to extend tests at lower level, let's do so.

akumetsu183:
> > Do we want to extend e2e tests with this new feature of sending custom attributes?
> 
> Loos like good idea to me as I do see test coverage only the config extraction, not the utilization - @akumetsu183 could you please extend the e2e test? Or if it is possible to extend tests at lower level, let's do so.

Sure thing. Extended.

jirkajanecek (approved):

adschr (approved):

PR #20, opened 2025-01-09
feature: Add k8s Distribution

simek-m:
#### Description
Build a new distribution based on https://github.com/solarwinds/swi-k8s-opentelemetry-collector.

#### Testing
The image for `k8s` distribution is built in the CI pipeline.


.github/workflows/buildAndDeploy.yml
```diff
@@ -103,39 +126,55 @@ jobs:
...
+#
+#      - name: Create Multi-arch Manifest for k8s Image
+#        run: |
+#          docker manifest create ${{ env.DOCKERHUB_IMAGE }}:${{ steps.get-tag.outputs.tag }}  \
+#            --amend ${{ env.DOCKERHUB_IMAGE }}@$(jq -r '.manifests[] | select(.platform.os == "linux" and .platform.architecture == "amd64") | .digest' manifest-k8s.json) \
+#            --amend ${{ env.DOCKERHUB_IMAGE }}@$(jq -r '.manifests[] | select(.platform.os == "linux" and .platform.architecture == "arm64") | .digest' manifest-k8s.json)
+#
+#      - name: Push Multi-arch Manifest for k8s Image
+#        run: |
+#          docker manifest push ${{ env.DOCKERHUB_IMAGE }}:${{ steps.get-tag.outputs.tag }}-k8s
```
akumetsu183, line 157:
```suggestion
#          docker manifest push ${{ env.DOCKERHUB_IMAGE }}:${{ steps.get-tag.outputs.tag }}-k8s
#          docker manifest push ${{ env.DOCKERHUB_IMAGE }}:latest-k8s
```

Just an idea.

akumetsu183:
Some readme update might be in order, to mention this new docker image.

.github/workflows/buildAndDeploy.yml
```diff
@@ -35,15 +35,20 @@ jobs:
...
-        run: > 
+      - name: Build and Test - Full Image
+        run: >
           docker build . --file build/docker/Dockerfile
-          --tag solarwinds-otel-collector:${{ steps.generate-tag.outputs.tag }}
-          --tag solarwinds-otel-collector:latest
+          --tag ${{ env.DOCKERHUB_IMAGE }}:${{ steps.generate-tag.outputs.tag }}
+          --tag ${{ env.DOCKERHUB_IMAGE }}:latest
+
+      - name: Build and Test - k8s Image
```
adschr, line 48: thinking about build times. This will be smaller distribution but still running it in sequence after the 'full' will make the jobs run for quite long time. And for the dockerhub publishing we would get into really long build times. So maybe for starters we could extract to GH composite actions (not sure we need different workflows yet, those are also an option) and have k8s stuff in separate jobs to enable parallel runs (as I don't think steps can be ran in parallel)? This will have an implication to the whole workflow, though (should I proceed publishing A when B failed) so let's discuss the options.

build/docker/Dockerfile.k8s
adschr, line 1: not sure what dockerfile will allow us. But wouldn't it be possible to adjust the original dockerfile with tags as ARG, and base this one on that, adding the k8s goodies as an extra? leaving that as a thought as there is not that much going on in original one, but I was just wondering on possibilities of reuse rather than C&P

.github/workflows/buildAndDeploy.yml
adschr: as discussed offline, let's do this in parallel, but drop the final image publishing part for now.

build/docker/Dockerfile.k8s
simek-m: @adschr 
- The `Dockerfile.k8s` can reference the local image build from `Dockerfile` and it can be parametrised using `ARG` to set a build tag when building the image. 
- The `Dockerfile.k8s` would need to have an `ARG` to parametrise the base image with the tag of the local image built during the build - like `FROM solarwinds-otel-collector:{VERSION}-k8s`.
- The `Dockerfile.k8s` adds `journactl` and `wrapper` and overrides `ENTRYPOINT` to `wrapper`.

Yes, I think that could work unless I'm missing something related to GitHub Actions environments. Personally, a simpler way could be move everything shared to the `Makefile` and use `ARG`s to parametrise version of images.

adschr: Sure. We can start with what is presented here and iterate over it in follow-up, defering the decision on which path to take there as we may be wiser in desired use then.

CONTRIBUTING.md
```diff
@@ -5,3 +5,22 @@ Thank you for your interest in contributing to SolarWinds OpenTelemetry Collecto
 Currently, this project is being developed by the SolarWinds team. 
 We are actively working on stabilizing its foundations and defining contribution guidelines.
 Once ready, we will update the `CONTRIBUTING.md` file with clear instructions.
+
+## Development
```
adschr, line 9: we can refer to docs/development-guidelines.md

Makefile
```diff
@@ -1,6 +1,6 @@
 include Makefile.Common
 
-ALL_SRC := $(shell find . \( -name "*.go" -o -name "*.sh" \) \
```
adschr, line 3: please make sure there is follow-up to reintroduce scripts check

adschr (approved):

akumetsu183 (approved):

adschr (approved):

akumetsu183 (approved):

adschr (approved):

akumetsu183 (approved):

PR #23, opened 2025-01-15
feature: Add version to uptime metric

david-sevcik:
#### Description

Adds version as `sw.otelcol.collector.version`  attribute to the uptime metric.

#### Testing

Start the collector with the solarwinds extension and correct settings and look, if the metric has this attribute

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -14,4 +14,4 @@
 
 package version
 
-const Version = "0.113.5"
+const Version = "0.113.6"


jirkajanecek (dismissed):
I would like to open a discussion about package versioning, when package is actually local and `replace` is applied.
- We apply `v0.113.0` for local packages. Personally, I would not expect to find replace to local place.
- My proposal would be `v0.113.0-local` to hold version "aligned", but point to its local replaced existence.

 I am open to any proposal to be aligned with Collector conventions.

adschr:
> I would like to open a discussion about package versioning, when package is actually local and `replace` is applied.
> 
> * We apply `v0.113.0` for local packages. Personally, I would not expect to find replace to local place.
> * My proposal would be `v0.113.0-local` to hold version "aligned", but point to its local replaced existence.
> 
> I am open to any proposal to be aligned with Collector conventions.

This is definitely good discussion point. Throughout the usages there is now misalignment of versions used. The _Collector way_ is to use the replace as we do, but refer to the component in up-to-date version. There are these deficiences on this repo's side for this approach:
 - components refer to hardcoded nonexistent versions
 - the tags (thus versions) are published for the collector itself, but not for its components (ie. see [contrib tags](https://github.com/open-telemetry/opentelemetry-collector-contrib/tags?after=receiver%2Fvcenterreceiver%2Fv0.117.0))

Both are set to be addressed in the follow-up that will deal with the centralized versioning across the repository - and from there we can decide how to evolve further in this regard.

adschr (dismissed):
LGTM - it just looks like licences need to be updated in the newly added files.

extension/solarwindsextension/internal/heartbeat.go
```diff
@@ -163,5 +164,6 @@ func (h *Heartbeat) decorateResourceAttributes(resource pcommon.Resource) error
 	if h.collectorName != "" {
 		resource.Attributes().PutStr("sw.otelcol.collector.name", h.collectorName)
 	}
+	resource.Attributes().PutStr("sw.otelcol.collector.version", version.Version)
```
adschr, line 167: can we extend e2e test to check we inclide the attribute in results?

adschr (approved):

jirkajanecek (approved):

jirkajanecek (approved):

pkg/version/version.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package version
+
+const Version = "0.113.5"
```
adschr, line 17: Note: We need to make sure this is bumped to 0.113.6 prior to the next (or current, according to changelog) release to make sure the versions are in-sync. The centralized versioning follow-up should hopefully make the process less dependent on the individual updates.

adschr (approved):

jirkajanecek (approved):

PR #25, opened 2025-01-16
feature: adding attribute without_entity

akumetsu183:
#### Description
Adding attribute without_entity to opt out of collector entity creation in solarwinds observability.

### Testing
Tested on branchbuild of entity service, both cases.

### DIFF
--- a/extension/solarwindsextension/README.md
+++ b/extension/solarwindsextension/README.md
@@ -40,7 +40,7 @@ extensions:
 - `data_center` (mandatory) - Data center is the region you picked during the sign-up process. You can easily see in URLs after logging in to SolarWinds Observability SaaS - it's either `na-01`, `na-02` or `eu-01`. Please refer to the [documentation](https://documentation.solarwinds.com/en/success_center/observability/content/system_requirements/endpoints.htm#Find) for details.
 - `collector_name` (mandatory) - The collector name passed in the heartbeat metric (as `sw.collector.name` resource attribute) to identify the collector. Doesn't have to be unique.
 - `resource` (optional) - You can specify additional attributes to be added to the `sw.otecol.uptime` metric. 
-- `without_entity` (optional) - You can disable Collector entity creation in Solarwinds Observability, by setting it to `true`. Otherwise, entity creation is enabled by default.
+- `without_entity` (optional) - You can disable Collector entity creation in SolarWinds Observability by setting it to `true`. If not configured, entity creation is enabled by default.
 ## Development
 - **Tests** can be executed with `make test`.
 - After changes to `metadata.yaml` generated files need to be re-generated with `make generate`. The [mdatagen](http://go.opentelemetry.io/collector/cmd/mdatagen) tool has to be in the `PATH`.
--- a/extension/solarwindsextension/README.md
+++ b/extension/solarwindsextension/README.md
@@ -40,6 +40,7 @@ extensions:
 - `data_center` (mandatory) - Data center is the region you picked during the sign-up process. You can easily see in URLs after logging in to SolarWinds Observability SaaS - it's either `na-01`, `na-02` or `eu-01`. Please refer to the [documentation](https://documentation.solarwinds.com/en/success_center/observability/content/system_requirements/endpoints.htm#Find) for details.
 - `collector_name` (mandatory) - The collector name passed in the heartbeat metric (as `sw.collector.name` resource attribute) to identify the collector. Doesn't have to be unique.
 - `resource` (optional) - You can specify additional attributes to be added to the `sw.otecol.uptime` metric. 
+- `without_entity` (optional) - You can disable Collector entity creation in Solarwinds Observability, by setting it to `true`. Otherwise, entity creation is enabled by default.
 ## Development
 - **Tests** can be executed with `make test`.
 - After changes to `metadata.yaml` generated files need to be re-generated with `make generate`. The [mdatagen](http://go.opentelemetry.io/collector/cmd/mdatagen) tool has to be in the `PATH`.
--- a/extension/solarwindsextension/config_test.go
+++ b/extension/solarwindsextension/config_test.go
@@ -47,7 +47,7 @@ func TestConfigUnmarshalFull(t *testing.T) {
 		IngestionToken:      "TOKEN",
 		CollectorName:       "test-collector",
 		Resource:            attributeMap,
-		WithoutEntity:       false,
+		WithoutEntity:       true,
 	}, cfg)
 }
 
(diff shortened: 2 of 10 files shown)


extension/solarwindsextension/README.md
```diff
@@ -40,6 +40,7 @@ extensions:
 - `data_center` (mandatory) - Data center is the region you picked during the sign-up process. You can easily see in URLs after logging in to SolarWinds Observability SaaS - it's either `na-01`, `na-02` or `eu-01`. Please refer to the [documentation](https://documentation.solarwinds.com/en/success_center/observability/content/system_requirements/endpoints.htm#Find) for details.
 - `collector_name` (mandatory) - The collector name passed in the heartbeat metric (as `sw.collector.name` resource attribute) to identify the collector. Doesn't have to be unique.
 - `resource` (optional) - You can specify additional attributes to be added to the `sw.otecol.uptime` metric. 
+- `without_entity` (optional) - You can disable Collector entity creation in Solarwinds Observability, by setting it to `true`. Otherwise, entity creation is enabled by default.
```
simek-m, line 43: Minor:
```suggestion
- `without_entity` (optional) - You can disable Collector entity creation in Solarwinds Observability by setting it to `true`. Otherwise, entity creation is enabled by default.
```
Should it be `SolarWinds Observability SaaS` to match other occurences?

extension/solarwindsextension/config_test.go
```diff
@@ -47,6 +47,7 @@ func TestConfigUnmarshalFull(t *testing.T) {
 		IngestionToken:      "TOKEN",
 		CollectorName:       "test-collector",
 		Resource:            attributeMap,
+		WithoutEntity:       false,
```
simek-m, line 50: This test should use `true` because `false` is the default, zero value of `bool` and it would pass even when not parsed from the config.

extension/solarwindsextension/testdata/without_entity.yaml
```diff
@@ -0,0 +1,4 @@
+token: "YOUR-INGESTION-TOKEN"
+data_center: "na-01"
+collector_name: "test-collector"
+without_entity: true
```
simek-m, line 4: Line break missing.

extension/solarwindsextension/config_test.go
simek-m: I think it can then replace the `TestConfigWithoutEntity` test.

internal/e2e/containers.go
```diff
@@ -204,6 +209,49 @@ func runGeneratorContainer(
...
 }
 
+// Starts the receiving collector, and the test collector.
+// Test collector container is started with the supplied config file to be tested.
+// Returns the receiving collector container instance, so tests can check the ingested data is as expected.
+func startCollectorContainers(
+	t *testing.T,
+	ctx context.Context,
+	config string,
+	signalType string,
```
simek-m, line 219: The `signalType` could be a pseudo-enum with a string method as an improvement:

```
type SignalType int

const (
    Logs SignalType = iota
    Metrics
// ...
)

func (s SignalType) String() string {
	switch s {
	case Logs:
		return "logs"
	case Metrics:
		return "metrics"
	case Traces:
		return "traces"
    default:
        panic("unexpected signal type")
	}
}
```

More examples [here](https://www.sohamkamani.com/golang/enums/)

internal/e2e/containers.go
```diff
@@ -204,6 +209,49 @@ func runGeneratorContainer(
...
+) testcontainers.Container {
+
+	net, err := network.New(ctx)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
akumetsu183: I think of it as a common test step.

I would leave it unless someone needs the errors, then they can rewrite it.

internal/e2e/containers.go
akumetsu183: Done.

extension/solarwindsextension/testdata/without_entity.yaml
akumetsu183: Removed, this test id gone.

extension/solarwindsextension/config_test.go
akumetsu183: Done.

extension/solarwindsextension/README.md
akumetsu183: Updated.

internal/e2e/containers.go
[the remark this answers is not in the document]
simek-m: I principally agree with David that returning an error and asserting in the test would be nice but if it's simpler this way I'm OK with it too.

You can mark this helper function with `t.Helper()`.

internal/e2e/containers.go
```diff
@@ -204,6 +209,49 @@ func runGeneratorContainer(
...
+		"--otlp-insecure",
+		"--otlp-endpoint", fmt.Sprintf("%s:%d", testedContainer, port),
+		"--otlp-attributes", fmt.Sprintf("%s=\"%s\"", resourceAttributeName, resourceAttributeValue),
+	}
+
+	gContainer, err := runGeneratorContainer(ctx, net.Name, cmd)
+	require.NoError(t, err)
+	testcontainers.CleanupContainer(t, gContainer)
+
+	<-time.After(collectorRunningPeriod)
```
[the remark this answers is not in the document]
simek-m: I'm not sure if the "sleep" is needed at all (this is what gets called):
https://github.com/testcontainers/testcontainers-go/blob/8b5211f88a6d4ce21164209ab004aa09f6bfde3f/docker.go#L226

The library is not heavy on documentation...

extension/solarwindsextension/testdata/full.yaml
```diff
@@ -5,3 +5,4 @@ endpoint_url_override: "127.0.0.1:1234"
 resource:
   att1: "custom_attribute_value_1"
   att2: "custom_attribute_value_2"
+without_entity: true
```
simek-m, line 8: Missing a new line.

simek-m (approved):
Great job!

extension/solarwindsextension/internal/heartbeat.go
```diff
@@ -165,6 +167,11 @@ func (h *Heartbeat) decorateResourceAttributes(resource pcommon.Resource) error
 	if h.collectorName != "" {
 		resource.Attributes().PutStr(CollectorNameAttribute, h.collectorName)
 	}
+	if h.withoutEntity {
+		resource.Attributes().PutStr("sw.otelcol.collector.entity_creation", "off")
+	} else {
+		resource.Attributes().PutStr("sw.otelcol.collector.entity_creation", "on")
+	}
```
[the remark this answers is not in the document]
akumetsu183: There is no ternary in go.

PR #30, opened 2025-01-24
feature: tag all signals with entity_creation attribute

akumetsu183:
Doing this to associate all outputs with entity

#### Description
Tagging all signals with entity_creation attribute.
Doing this in order for all collector outputs to be properly associated with created entity in SolarWinds Observability.

When without_entity: true is set in extension configuration, attribute is not added to any signals (and also removed from uptime heartbeat metric)

#### Testing
Manual testing against SWO

### DIFF
--- a/exporter/solarwindsexporter/solarwinds_exporter.go
+++ b/exporter/solarwindsexporter/solarwinds_exporter.go
@@ -39,6 +39,7 @@ const (
 
 var (
 	ErrSwiExtensionNotFound = errors.New("solarwinds extension not found")
+	EntityCreationOnValue   = "on"
 )
 
 type solarwindsExporter struct {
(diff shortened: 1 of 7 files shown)


exporter/solarwindsexporter/solarwinds_exporter.go
```diff
@@ -237,6 +243,10 @@ func (swiExporter *solarwindsExporter) pushMetrics(ctx context.Context, metrics
 			solarwindsextension.CollectorNameAttribute,
 			swiExporter.config.collectorName,
 		)
+		resource.Attributes().PutStr(
```
adschr, line 246: in the sense of the 'opt in' setting to disable entity creation, can we add the attribute only in case we want to infer the entity? The `entityCreation` string prop could be changed to simple bool wrote-through the setting and re attributes would be updated only in case we want to add the entity... or is there any other reason to mark the signals for both cases?

akumetsu183: I only did it to keep the behavior the same with the uptime metric. Wchich also sends "on" or "off".
But if we want to only send "on" and drop it otherwise, then I would drop it from everything, including the heartbeat/uptime metric.

akumetsu183: Discussed on SU. I'll update this to only send create_entity: "on", and otherwise not include the atrribute if user opts out.

adschr: I see, I didn't participate in the original PR so the uptime behavior slipped my radar. Sending the required attribute only when needed uniformly sounds good.

adschr (approved):

PR #33, opened 2025-01-29
chore: Add Windows Architecture

gantrior:
* Adding windows build to the pipeline and to the multiarch
* Add possibility to run collector as windows service
* Since windows build takes a long time to run (30-40 minutes), so run it only on main and release branches to have some continuous check that build on windows still works. 
* Windows architectures are being added to multi-arch manifest
* I enabled run also on manual runs, here is manual run I triggered for this job to show that windows build is fine: https://github.com/solarwinds/solarwinds-otel-collector/actions/runs/13054997673
* NOTE: I couldn't test if the publishing part works
* This is prerequisite for: https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/879


jirkajanecek (approved):
As far as I am able to say...
...it looks good to me.

PR #40, opened 2025-02-10
chore: Update build pipeline

pstranak-sw:
#### Description
Create new build pipeline for building and deploying images.

#### Testing
This will need to be tested by actually running the pipeline.


.gitattributes
pstranak-sw, line 1: Added .gitattributes to ensure that scripts have proper EOLs regardless of OS where the repo is cloned.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 
+## vNext
```
pstranak-sw, line 3: The `vNext` title is an anchor looked for by the release script. The script replaces it with an actual version.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,188 @@
...
+      image_tag: ${{ steps.get-tag.outputs.value }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Get Current Tag
+        id: get-tag
+        run: echo "tag=v$(grep -oP '(?<=const Version = ")[^"]+' "../../pkg/version/version.go")" >> $GITHUB_OUTPUT
+
+  build_and_test:
+    uses: ./.github/workflows/buildAndTest.yml
```
pstranak-sw, line 29: The `Deploy Images` workflow (executed when merging new image version to `main`/`release`) depends on the `Build and Test Images` workflow. The latter should be executed in whole, including Windows images.

.github/workflows/checkPullRequestSemantics.yml
```diff
@@ -20,6 +20,6 @@ jobs:
         PR_TITLE: ${{ github.event.pull_request.title }}
       run: |
         if [[ ! ${PR_TITLE} =~ ^(build|ci|chore|docs|feature|fix|perf|refactor|style|test):[[:space:]](#[[:digit:]]+[[:space:]]){0,1}([^ ,#].*)$ ]]; then
-          echo "::error::Invalid Pull Request title. Please update it according to guidelines: https://github.com/solarwinds/solarwinds-otel-collector/docs/development-guidelines.md#pull-requests"
+          echo "::error::Invalid Pull Request title. Please update it according to guidelines: https://github.com/solarwinds/solarwinds-otel-collector/blob/main/docs/development-guidelines.md#pull-requests"
```
pstranak-sw, line 23: The link did not work.

.github/workflows/buildAndDeploy.yml
pstranak-sw, line 1: This one was replaced by `.github/workflows/buildAndTest.yml` and `.github/workflows/deploy.yml`.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,188 @@
...
+  get_current_tag:
+    runs-on: ubuntu-latest
+    outputs:
+      image_tag: ${{ steps.get-tag.outputs.value }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Get Current Tag
+        id: get-tag
+        run: echo "tag=v$(grep -oP '(?<=const Version = ")[^"]+' "../../pkg/version/version.go")" >> $GITHUB_OUTPUT
```
adschr, line 26: could we also add step that would use the version to run `make prepare-release` and check git diff (fail in case there are some changed files)? Just as a failsafe that sb doesn't just change the version file, but does not reflect changing other places as well.

pstranak-sw: Good point. Added

adschr (approved):

PR #46, opened 2025-02-12
chore: #56 Add docker scout action

janatrelakova:
#### Description
Introducing Docker scout to collector build pipeline.

- `compare` docker scout action -> on regular PR `build_and_test`: docker scout comment with link is added, the link will redirect to build summary
  - `compare` gh action provides writing a comment but as we have many images to compare new build kept overwritting the existing comments, so we decided to part from that, there is issue for this particular bug https://github.com/docker/scout-action/issues/52
  - "Github run ID" is being used for comment identification as it is not possible to assign specific ID or an attribute to new comment that would make it possible to easily identify it in case of updating from GHA
  - this is the output 
image
image

- `cves` docker scout action -> after merge + manual trigger of the windows pipeline, we should see some files/outputs in the security tab, if it is not there after everything is green there may be something additional needed on SRE side
  - linux is currently ignored due to complicated build pipeline that will be aligned to windows style in follow-up
  - only `critical` and `high` severities are being shown -> this creates incosistency with `compare` action in the summary, but we can take care of medium and less important vulnerabilities and align these settings
  - `cves` action also provides summary in the build output -> after workflow is triggered and completed summary is available 
  - sarif files are published after build is completed

#### Testing
No test cases were added as this is github configuration.

### Merge scenario
- [x] resolve comments
- [x] ignore release check
- [x] reapproval
- [x] merge


janatrelakova:
> ## 🔍 Vulnerabilities of `solarwinds-otel-collector:v131-c98fc02-k8s`
> 📦 Image Reference `solarwinds-otel-collector:v131-c98fc02-k8s`
> digest	`sha256:461751717b1b860047b8cb59c2b9d76498309cd647af4b2d51faeeac0507a954`
> vulnerabilities	critical: 0 high: 0 medium: 0 low: 0
> platform	linux/amd64
> size	55 MB
> packages	336
> 📦 Base Image `gcr.io/distroless/static:nonroot`

This comment will not be visible, as it is product of cves analysis that is NOT run on PR but after manual trigger of workflow.

.github/workflows/deploy.yml
```diff
@@ -134,6 +134,34 @@ jobs:
...
+          token: ${{ secrets.ENOPS5919_DOCKER_SCOUT_CI_PAT }}
+          tag: ${{ steps.generate-tag.outputs.tag }}-nanoserver-ltsc2022-k8s
+
+      - name: Analyze cves Windows 2019 k8s
+        uses: ./.github/actions/analyzeCves
+        with:
+          username: ${{ vars.ENOPS5919_DOCKER_SCOUT_CI_USER }}
+          token: ${{ secrets.ENOPS5919_DOCKER_SCOUT_CI_PAT }}
+          tag: ${{ steps.generate-tag.outputs.tag }}-nanoserver-ltsc2019-k8s
+
```
janatrelakova, line 164: I would like to have this in separate workflow, but with new runner there are not docker images. I could resolve it by doing `docker load` twice but I don't consider that as the best solution due to computational requirements.

.github/actions/analyzeCves/action.yml
```diff
@@ -0,0 +1,34 @@
+name: Docker scout CVE analysis
+description: Analyse CVEs with Docker Scout and store it as an artifact
+
+inputs:
+  username:
+    description: Username for logging into the Docker Scout
+    required: false
```
adschr, line 7: username and token aren't utilized directly - is their use automagical? And is it expected for these to be optional?

.github/actions/dockerScoutComment/action.yml
```diff
@@ -0,0 +1,49 @@
...
+            issue_number: context.issue.number,
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            body: "${{ steps.set_comment_body.outputs.comment_body }}"
+          })
+
+    - name: Update existing comment
+      if: ${{ steps.find_comment.outputs.comment_id != '' }}
+      shell: bash
+      run: |
```
adschr, line 44: using script won't work?

```
        github.rest.issues.updateComment({
            owner: context.repo.owner,
            repo: context.repo.repo,
            comment_id:  "${{ steps.find_comment.outputs.comment_id }}",
            body: "${{ steps.set_comment_body.outputs.comment_body }}"
        })
```

.github/actions/dockerScoutScan/action.yml
```diff
@@ -0,0 +1,57 @@
...
+    description: Image to be compared
+    required: true
+  current-tag:
+    description: Tag of the image to be compared
+    required: true
+  compare-to-image:
+    description: Image name to compare to
+    required: true
+  compare-to-tag:
+    description: Image to compare to, required if `compare-to-regex` is not provided, if both provided, this is prioritized
```
adschr, line 24:
```suggestion
    description: Tag to compare to, required if `compare-to-regex` is not provided, if both provided, this is prioritized
```

.github/actions/dockerScoutScan/action.yml
```diff
@@ -0,0 +1,57 @@
...
+      shell: bash
+      run: exit 1
+
+    - name: Log into registry with PAT to use SCOUT.
+      uses: docker/login-action@v3
+      with:
+        username: ${{ inputs.username }}
+        password: ${{ inputs.token }}
+
+    - name: Get latest tag
```
adschr, line 44:
```suggestion
    - name: Get tag by regex
```
Maybe? As there is no explicit condition on that tag being also the latest.

adschr: ultranit: I'd also consider naming the input `compare-to-regex-tag` or similar. With just 'regex' i get the notion the full name will be checked by regex (as the competing input is named `compare-to-tag`)..

adschr: another thought, bound to happen in potential follow-up. We cannot use version.go as for deployment it will have updated and unreleased version within, but at this point we should be able to say what was the latest successful release (from GH release/tag).. each step could use that info and just append the specific distribution string. This way we could get rid of querying all the tags (repeatedly, for each step utilizing this action, I presume) and manual filtering. 

(Of course this hypothesis works with our belief in the deployment working as it should.. but I presume that is where we want to head anyway).

.github/workflows/deploy.yml
adschr: discussed offline: there are not many options at this point. Steps on their own can't be ran in parallel (nor would it be possible due to transitive dependency on external action that has this limitation as well), extracting to separate job would invoke the need to download the sizable artifacts yet another time.
Let's go with what we have and evaluate the footprint with actual hard data from the run.

.github/actions/dockerScoutComment/action.yml
janatrelakova: It works, thanks.

PR #58, opened 2025-02-26
chore: Use pdatatest package in some E2E tests to compare received OTEL messages

pstranak-sw:
#### Description
Refactor some of the testing code to reduce code duplication.

And for some test, start using [pdatatest](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/pdatatest) for comparing received test data with expected data stored in committed files.

The expected data is in folder `internal\e2e\testdata\expected`. Right now, it's a dump of OTEL messages serialized into JSON, one message per line. It's the same format that is logged when calling helper function `loadResultFile` in the E2E tests.

In general, when considering comparing received OTEL messages with static expected data using the `pdatatest` package, one should consider the following points:
* It’s built for comparing single, well defined OTEL messages, ideally created manually in test or using the `telemetrygen` tool.
* Some of our E2E tests are for receivers, which generate metrics with many values/attributes that change every time. Those are bad candidates for using `pdatatest`. On the other hand, tests for processors or exporters are a good fit.
* Heartbeat is mixed within the OTEL messages, so the test has to account for it.
* OTEL messages are received in random order, so comparing more than one requires additional code in the test.
* OTEL messages may contain SolarWinds OTEL Collector version in them, which changes over time. Again, the tests have to account for that.


jirkajanecek (approved):
I like it.

PR #61, opened 2025-02-26
chore: Removing wrapper as it is no longer needed

gantrior:
#### Description
Removing wrapper as it was already removed from K8s image in https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/919

fsync is an option on filestorage since v0.96.0 (see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/20266) and should prevent corruption of the storage which was the reason why we did wrapper in the first place.

#### Testing
Tested in swi-k8s-opentelemetry-collector image

### DIFF
--- a/.github/actions/analyzeCves/action.yml
+++ b/.github/actions/analyzeCves/action.yml
@@ -14,11 +14,9 @@ runs:
       uses: docker/scout-action@v1
       with:
         command: cves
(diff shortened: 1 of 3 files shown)


adschr (approved):
LGTM

Issue #72, opened 2025-03-03
Migrate `k8sobjectsreceiver` from `swi-k8s-opentelemetry-collector repository`


mpecenka:
### Is your feature request related to a problem? Please describe.

There is `k8sobjectsreceiver` that is needed by k8s distribution and is currently located at https://github.com/solarwinds/swi-k8s-opentelemetry-collector/tree/master/src/receiver/k8sobjectsreceiver

### Describe the solution you'd like

Add new receiver called `swok8sobjectsreceiver` and migrate code as-is

PR #74, opened 2025-03-05
feature: #72 modify swok8sobjectsreceiver

mpecenka:
Moving modified k8sobjectsreceiver from https://github.com/solarwinds/swi-k8s-opentelemetry-collector

The SWO Kubernetes Objects Receiver (swok8sobjects) is based on the existing [k8sobjectsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sobjectsreceiver) from OpenTelemetry. In addition to the original functionality, it adds attributes to the log messages indicating which part of the manifest was changed (sw.metadata.changed, sw.status.changed, sw.spec.changed).

Additionally, this receiver can be configured to use storage, allowing it to persist data across collector restarts.


jirkajanecek (approved):
Looks fine.

Issue #76, opened 2025-03-05
Improve GHA pipeline


janatrelakova:
### Is your suggestion related to a general technical or maintenance issue? Please describe.

- Cut ARM build time by using native ARM machine, not QEMU
- Refine set of workflows triggered on push to main, on PR change, on deploy
- Developer should be able to trigger Windows machine

### Describe the solution you'd like

- ARM build takes shorter period of time
- Reuse as much as possible, but think about readability

PR #77, opened 2025-03-05
chore: #76 Improve GHA pipeline

janatrelakova:
#### Description
- To speed up ARM build we could go with cross-compilation in Dockerfiles or running it on separate runners (arm/amd/windows machines)
- Considering we will probably migrate to goreleaser I chose to go with the ARM native machine as discussed before
- Note: opentelemtry-collector uses matrix build but for cross-compilation and opentelemtry-collector-releases uses goreleaser
- Goreleaser uses QEMU + `buildx` builder which is the thing we wanted to remove (happening in this PR)
  - but I chose arm machine so we can ran goreleaser on different runners so we don't have to use `buildx` which makes it really slow (100 minutes vs 8 minutes for basic arm image)
 
**Tracking Issue:** https://github.com/solarwinds/solarwinds-otel-collector/issues/76

#### Testing
Github configuration -> no test cases.

#### Notes
- e2e tests are runnable only on latest tag -> IMHO we should create follow-up for that
- unit tests are run only in Dockerfile -> it takes time, IMHO follow-up for that would be great,  if we want to run tests on all images we are deploying
- open-telemetry naming of workflow files is kebab case not camel case, kebab case seems more readable to me
- I did not delete any of file but when we are happy with the new workflows the old ones should be deleted in this PR

### Concept
- on `PR` run `build_linux_amd`, it will run basic and k8s distribution -> discussed with k8s, that it would be better to see implicitly on the PR the state -> but could be removed for sure
- on `merge to main - change to version.go` run `build_and_test`
- on `merge to main + change to version.go run `deploy`, just deploy because it will run the buiald and test on all images inside
- things I am not sure about
  - merge linux workflows to one 
  - separate `build_linux_amd` to `build_linux_amd` + `run_on_pr.yml`, so we can reuse amd workflow better 🤷🏼‍♀️ 
  - I don't like docker scout composite action it takes too much parameters, but I don't think using workflow would help that much

#### TODO:
- documentation of pipeline, added comments was to 
- TODO in deploy workflow (mainly copying existing steps to the new workflow `new_deploy.yml`) I did not proceed on that as some slight adjustments like names could be needed
- uploading artifact with correct name so it will be recognizable when downloading artifact


adschr:
**Closed:** to be superseded by final release pipeline.

PR #78, opened 2025-03-07
chore: improved dependabot configuration

akumetsu183:
#### Description
Extending dependabot configuration


jirkajanecek (dismissed):

.github/dependabot.yaml
```diff
@@ -19,32 +20,117 @@ updates:
...
+        - "*opentelemetry*"
+        - "*solarwinds-otel-collector*"
+      collector-miscellaneous:
+        exclude-patterns:
+          - "*opentelemetry*"
+          - "*solarwinds-otel-collector*"
+    commit-message:
+      prefix: "chore: "
+
+  - package-ecosystem: gomod
```
adschr, line 49: a though about the bump benefit: it should mainly allow us 'one-peek' solution to have overview how smooth bumping the otel to new version should be. So from this POV what I was thinking is having an overall setup for all the opentelemetry deps usage throughout the repository as when we bump, we want to bump everywhere - that is what I would expect from manual PR.

Having the settings separated by components I am afraid of two things:
- definitely, adding setup for new component will be forgotten at some point
- lessened value added of the bump, as as a result we would have one PR for collector + N PRs per components (if I understand it correctly)

Would it be possible to have some general bump job instead? We might add per-component granularity when that need arises, until then I'd go with more straightforward definition (& consequently straightforward use for maintenance purposes).

But maybe I'm missing something :)

akumetsu183: There are pros and cons.

When separated, there is a higher chance that some things might auto pass and could be just merged. And any areas in need o manual resolution would be more isolated. 

But that has the drawbacks you mentioned. 

I'll update this to be one big group.

.github/dependabot.yaml
```diff
@@ -19,32 +20,33 @@ updates:
...
-    ignore:
-      - dependency-name: "github.com/open-telemetry/opentelemetry-collector-contrib*"
-      - dependency-name: "go.opentelemetry.io/collector*"
-    # Disable version updates
-    open-pull-requests-limit: 0
+      - dependency-type: "direct"
     groups:
-      minor-and-patch:
-        applies-to: security-updates
+      go-dependencies:
```
adschr, line 48: could we possibly leave two groups:
- otel
- security updates (as it was before in the `minor-and-patch`)

PR #103, opened 2025-03-17
chore: diff-generator-tool

akumetsu183:
#### Description
 diff-generator-tool

Allows us to get release notes information about breaking changes, deprecations and enhancements only on components that we use. Returns it as a text ouput.

Example command.

`go run ./main.go --old v0.119.0 --new v0.121.0 --goModPath ./../../../cmd/solarwinds-otel-collector/go.mod --dependencyFilter opentelemetry-collector-contrib --repo opentelemetry-collector-contrib`

Ouptut, when places inside a github PR comment.
image


internal/tools/changes_analyzer/README.md
```diff
@@ -0,0 +1,42 @@
+# OpenTelemetry Changes Analyzer
```
akumetsu183, line 1: Where should I place this in the repo? 

This might not be the right place.

akumetsu183: Moved to /cmd

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+			blocks = append(blocks, componentBlock.String())
+		}
+	}
+	// separator between components
+	return strings.Join(blocks, "\n---\n")
+}
+
+// getMessage generates a formatted github formated message listing component changes between two versions. Optionally, encodes to base64.
+func getMessage(oldTag, newTag string, componentsOfInterest []string, opentelemetryRepo string, encode bool) (string, error) {
+	compareURL := fmt.Sprintf("https://github.com/open-telemetry/%s/compare/v%s...v%s", opentelemetryRepo, oldTag, newTag)
```
adschr, line 346: nit: compareURL is not utilized until after the next err check is passed, maybe I'd move it to relevant code section

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+				sectionMap[category] = append(sectionMap[category], changes...)
+				break // Assume only one category per heading
+			}
+		}
+	})
+	return sectionMap, nil
+}
+
+// getComponentChanges retrieves breaking changes, deprecations, and enhancements for specified components across versions.
+func getComponentChanges(versionOld, versionNew string, ourComponents []string, opentelemetryRepo string) (map[string]map[string][]string, error) {
```
adschr, line 205:
```suggestion
func getComponentChanges(versionOld, versionNew string, componentsOfInterest []string, opentelemetryRepo string) (map[string]map[string][]string, error) {
```
to stay consistent

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+
+		for _, rel := range releases {
+			if !rel.Prerelease {
+				allReleases = append(allReleases, rel.TagName)
+			}
+		}
+
+		linkHeader := resp.Header.Get("Link")
+		if linkHeader == "" {
+			url = ""
```
adschr, line 89: somehow breaking the loop seems more natural to me to clearly state 'this is where it ends' ... but I get the intent to finish the loop by completing the for condition.. and having it in condition marking the url a point of interest for the loop. Leaving as a gray-zone consideration.

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+	newVer, err := parseVersion(newVersion)
+	if err != nil {
+		return nil, fmt.Errorf("invalid new version %s: %v", newVersion, err)
+	}
+
+	// Filter versions within the range
+	var filtered []string
+	for _, tag := range allReleases {
+		ver, err := parseVersion(tag)
```
adschr, line 109: shouldn't we end with parsing error (being an exceptional state)?

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+		if err == nil && ver.GreaterThanOrEqual(oldVer) && ver.LessThanOrEqual(newVer) {
+			filtered = append(filtered, tag)
```
adschr, line 111: would it make sense to store versions in the `filtered` instead of strings? we could maybe omit the re-parsing in the slice sorting below

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+		vj, _ := parseVersion(filtered[j])
+		return vi.LessThan(vj)
+	})
+
+	return filtered, nil
+}
+
+// fetchReleaseNotes retrieves the HTML content of release notes for a specific version.
+func fetchReleaseNotes(version string, opentelemetryRepo string) (string, error) {
+	url := fmt.Sprintf("https://github.com/open-telemetry/%s/releases/tag/%s", opentelemetryRepo, version)
```
adschr, line 127: the part up to status check seems to be very similar to what happens for each url in `getVersionsBetween`. could we maybe extract (returninng the body and error)?

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+	if err != nil {
+		fmt.Printf("failed to read release notes body: %v", err)
+		return "", fmt.Errorf("failed to read release notes body: %v", err)
+	}
+	return string(body), nil
+}
+
+// extractReleaseSections extracts specified sections (e.g., Breaking changes, Deprecations, Enhancements) from HTML content.
+func extractReleaseSections(htmlContent string, sectionPhrases map[string]string) (map[string][]string, error) {
+	doc, err := goquery.NewDocumentFromReader(strings.NewReader(htmlContent))
```
adschr, line 150: there is no native package variant I presume? This package should be safe, though, right?

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		fmt.Printf("failed to read release notes body: %v", err)
+		return "", fmt.Errorf("failed to read release notes body: %v", err)
+	}
+	return string(body), nil
+}
+
+// extractReleaseSections extracts specified sections (e.g., Breaking changes, Deprecations, Enhancements) from HTML content.
+func extractReleaseSections(htmlContent string, sectionPhrases map[string]string) (map[string][]string, error) {
```
adschr, line 149: is there reason for having the incoming section phrases as `map[string]string`? Wouldn't it be possible just to input the sections of interest and deal with the map internally in this function instead (if it is not exposed nor intended to work-with outside of the function)?

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+				for node := s.Next(); node.Length() > 0; node = node.Next() {
+					if node.Is("h1, h2, h3") {
+						break
+					}
+					changeText := strings.TrimSpace(node.Text())
+					if changeText != "" {
+						lines := strings.Split(changeText, "\n")
+						for _, line := range lines {
+							trimmed := strings.TrimSpace(line)
+							if trimmed != "" {
```
adschr, line 178: dtto as comment above

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+								if strings.Contains(trimmed, ":") {
```
adschr, line 179: looking at the release changelogs, isn't it possible to accumulate the bullet points (HTML elements i presume) texts from the given sections? Or are there some specific cases that require this more intricate transformations.

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+	sort.Strings(components) // Sort components alphabetically
+
+	prPattern := regexp.MustCompile(`#(\d+)`) // Matches # followed by digits
+
+	for _, component := range components {
+		var componentBlock strings.Builder
+		componentBlock.WriteString(fmt.Sprintf("#### %s\n", component))
+
+		categories := componentChanges[component]
+		for _, category := range []string{"breaking_changes", "deprecations", "enhancements"} {
```
adschr, line 311: the categories of interest are used throughout functions. Maybe we could extract those and work with the array in relevant places?

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+							fmt.Sprintf("%s: %s", ver, change),
+						)
+					}
+				}
+			}
+		}
+	}
+
+	// Filter out components with no changes in any category
+	result := make(map[string]map[string][]string)
```
adschr, line 261: nit: the anonymous uber mapping might look a bit intimidating. would it be possible to define type alias for better readability? If it makes sense on more occassions.
```
type categoryToChangesMap = map[string][]string
```

adschr:
Grand amount of work, output looks awesome 👏  left some stylistic notes for consideration, we can surely go with the outputs for V1. A couple of general notes:
- would it be possible to somehow slice the `main.go`? maybe extract some formatting or data fetching functionality to separate files?
- would it be possible to add some general integration test? 😬 (or maybe in follow-up). Data fetching would need to be wrapped. My only concern is that this PR presents a lot of data handling and we should have some early feedback pipeline to alert us when some changes are to be done.. as this is also a public module, potentially affecting outputs for any of the consumers

cmd/changes-analyzer/main.go
akumetsu183: Done.

cmd/changes-analyzer/main.go
akumetsu183: Abstracted

cmd/changes-analyzer/main.go
akumetsu183: Rewritten, now getting it based on the <li> elements.

cmd/changes-analyzer/main.go
akumetsu183: Rewritten.

cmd/changes-analyzer/main.go
```diff
@@ -0,0 +1,449 @@
...
+		// Process section headers regardless of which changelog they belong to
+		for phrase, category := range sectionPhrases {
+			if strings.Contains(text, phrase) {
+				var changes []string
+				for node := s.Next(); node.Length() > 0; node = node.Next() {
+					if node.Is("h1, h2, h3") {
+						break
+					}
+					changeText := strings.TrimSpace(node.Text())
+					if changeText != "" {
```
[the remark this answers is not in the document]
akumetsu183: Rewritten.

cmd/changes-analyzer/main.go
akumetsu183: Internalized.

cmd/changes-analyzer/main.go
akumetsu183: Abstracted into getResponse

cmd/changes-analyzer/main.go
akumetsu183: Done.

cmd/changes-analyzer/main.go
akumetsu183: Updated.

cmd/changes-analyzer/main.go
akumetsu183: Inverted.

cmd/changes-analyzer/main.go
akumetsu183: Done.

cmd/changes-analyzer/main.go
akumetsu183: Moved.

adschr (dismissed):
looks good for v1! Just noticed tests seem to fail in the build 🤔

adschr (dismissed):

PR #120, opened 2025-04-04
Chore: fix dependabot issue with gonum

akumetsu183:
#### Description
Dependabot is failing because it gets 404s on https://gonum.org/v1/gonum?go-get=1.
This is a workaround.


.github/dependabot.yaml
```diff
@@ -14,12 +14,12 @@ updates:
...
     directory: /build/docker
-    labels:
-      - dependencies
-      - docker
     schedule:
       interval: weekly
       day: saturday
+    labels:
+      - dependencies
+      - docker
```
akumetsu183, line 22: Just to trigger dependabot.

Issue #127, opened 2025-04-14
SolarWinds Entity Connector


janatrelakova:
### Is your feature request related to a problem? Please describe.

Introduce new component, that will handle updates of entities and relationships.

### Describe the solution you'd like

The component should be connector to cover metrics and logs at the same time.

PR #128, opened 2025-04-17
feature: #127 Send update log for entity

janatrelakova:
# Description
Entity state update will be sent to SWO on every receive action of metrics and logs

## Notes
- Entity state update will be sent to SWO on every receive action of metrics and logs
  - I added it there so I don't have unused functions and don't have to comment out things 
- ADR schema has been changed, because these event logs does not work for properties without `@telemetryMapping` directive in entity schemas. 
  - Thus, we could remove `resource_attribute` and `property` from schema and make it plain list (`field` will not not be supported - as one can fulfill this requirement by using transform processor or other). 
  - New schema:
 ```
schema:
  entities:
    - entity: Snowflake
       id:
         - host.id
         - sw.something
       attributes:
         - host.name
 ```
- Currently, the timestamp is set to current time
  - it could make sense to fill it with timestamp of incoming telemetry
    - I would like to remind of one limitation of ingestion: if the timestamp is too old (outside of accepted interval) the telemetry is skipped -> however, it should be our problem as the polled timestamp of incoming telemetry should be relatively fresh
    - However, I would do the setting later, when it is more clear how using of OTTL context will look like
- The creation of `solarwindsentity` struct seems duplicated in factory, but I would wait for OTTL context and what it will bring to the picture
- 💡 When no events (logRecords) are found I would skip sending anything in `ConsumeLogs`, if not skipped it would send `{}` (`plog.NewLogs()`), now I just skip the `ConsumeLogs` so nothing is going out from the connector
- I wanted to decouple configuration schema from internal entities, it could look slightly duplicated, but I think they should be separated as schema representation could change and it would break down internal things

<details>

<summary>Example of the output log, see more details on confluence page</summary>

```
{
    "resourceLogs": [
        {
            "resource": {},
            "scopeLogs": [
                {
                    "scope": {
                        "attributes": [
                            {
                                "key": "otel.entity.event_as_log",
                                "value": {
                                    "boolValue": true
                                }
                            }
                        ]
                    },
                    "logRecords": [
                        {
                            "observedTimeUnixNano": "1744882253443874000",
                            "body": {},
                            "attributes": [
                                {
                                    "key": "otel.entity.id",
                                    "value": {
                                        "kvlistValue": {
                                            "values": [
                                                {
                                                    "key": "hashKeyOrder property mapped in ESS",
                                                    "value": {
                                                        "stringValue": "value"
                                                    }
                                                },
                                                {
                                                    "key": "hashKeyOrder property mapped in ESS",
                                                    "value": {
                                                        "stringValue": "value"
                                                    }
                                                }
                                            ]
                                        }
                                    }
                                },
                                {
                                    "key": "otel.entity.event.type",
                                    "value": {
                                        "stringValue": "entity_state"
                                    }
                                },
                                {
                                    "key": "otel.entity.type",
                                    "value": {
                                        "stringValue": "entity type as defined in ESS"
                                    }
                                },
                                {
                                    "key": "otel.entity.attributes",
                                    "value": {
                                        "kvlistValue": {
                                            "values": [
                                                {
                                                    "key": "attribute that is mapped in ESS, will update the state of entity",
                                                    "value": {
                                                        "stringValue": "value"
                                                    }
                                                }
                                            ]
                                        }
                                    }
                                }
                            ],
                            "traceId": "",
                            "spanId": ""
                        }
                    ]
                }
            ]
        }
    ]
}
```
</details>

**Tracking Issue:** https://github.com/solarwinds/solarwinds-otel-collector-releases/issues/127

# Testing
The following config and python script can be used to test the added functionality. 
Together they will create snowflake entities. Note that that is the only entity type supported now, as it is hardcoded in this partial implementation.

<details>
<summary>Example setting of connector</summary>

```
receivers:
  otlp:
    protocols:
      http:
        endpoint: 0.0.0.0:4318

extensions:
  solarwinds:
    token: ${TOKEN}
    data_center: "na-01"
    collector_name: ${COLLECTOR_NAME}
    endpoint_url_override: ${OVERRIDE_URL}
    without_entity: false

exporters:
  solarwinds:

connectors:
  solarwindsentity:
    schema:
      entities:
        - entity: Snowflake
          id:
            - host.id
            - sw.uams.plugin.instance.id
          attributes:
            - host.name
            
service:
  extensions: [solarwinds]
  pipelines:
    metrics/dev:
      receivers: [otlp]
      exporters: [solarwindsentity]
    logs/dev:
      receivers: [otlp]
      exporters: [solarwindsentity]
    logs/connector:
      receivers: [solarwindsentity]
      exporters: [solarwinds]
      
```
</details>

<details>
<summary>Script to simulate some logs/metrics. </summary>

```
import time
from typing import Dict
import requests

otlp_http_endpoint = 'http://localhost:4318/v1/'


def get_data(logs: bool) -> Dict[str, any]:
    """Returns mock metric data."""
    current_time = time.time_ns()
    metric_value = 10
    metric_name = "otlp.http.custom.metric"

    if logs:
        return \
        {
            "resource_logs": [
                {
                    "resource": {
                        "attributes": [
                            {
                                "key": "host.name",
                                "value": {
                                    "stringValue": "milan-host-name-4"
                                }
                            },
                            {
                                "key": "sw.uams.plugin.instance.id",
                                "value": {
                                    "stringValue": "milan-instance-id-4"
                                }
                            },
                            {
                                "key": "host.id",
                                "value": {
                                    "stringValue": "milan-host-id-4"
                                }
                            }
                        ],
                    },
                    "scope_logs": [
                        {
                            "log_records": [
                                {
                                    "observedTimeUnixNano": current_time,
                                    "body": { "value": { "stringValue": "aaa"}},
                                }
                            ]
                        }
                    ]
                }
            ]
        }
    # METRICS
    return \
{
    "resource_metrics": [
        {
            "resource": {
                "attributes": [
                    {
                        "key": "host.name",
                        "value": {
                            "stringValue": "milan-host-name-3"
                        }
                    },
                    {
                        "key": "sw.uams.plugin.instance.id",
                        "value": {
                            "stringValue": "milan-instance-id-3"
                        }
                    },
                    {
                        "key": "host.id",
                        "value": {
                            "stringValue": "milan-host-id-3"
                        }
                    }
                ],
            },
            "scope_metrics": [
                {
                    "metrics": [
                        {
                            "name": metric_name,
                            "description": "Have a bloody good day mate!",
                            "unit": "absolute unit",
                            "gauge": {
                                "data_points": [
                                    {
                                        "start_time_unix_nano": current_time,
                                        "time_unix_nano": current_time,
                                        "as_double": metric_value,
                                        "as_int": metric_value,
                                        "flags": 100
                                    }
                                ]
                            }
                        }
                    ]
                }
            ]
        }
    ]
}

switch = True

if switch:
    otlp_http_endpoint = f"{otlp_http_endpoint}logs"
else:
    otlp_http_endpoint = f"{otlp_http_endpoint}metrics"

for i in range(10):
    try:

        print(get_data(logs=switch))
        response = requests.post(otlp_http_endpoint, json=get_data(logs=switch))
        response.raise_for_status()  # Raise an exception for any HTTP errors
        print('POST request successful!', flush=True)
        time.sleep(10)
    except requests.exceptions.RequestException as e:
        print('POST request failed:', e, flush=True)
        time.sleep(10)

```

</details>

image


connector/solarwindsentityconnector/internal/log_builder.go
```diff
@@ -0,0 +1,110 @@
...
+// It returns true if the attribute was found and copied, false otherwise.
+func putAttribute(dest *pcommon.Map, key string, src *pcommon.Map) bool {
+	attrVal, ok := src.Get(key)
+
+	if !ok {
+		zap.L().Warn("attribute not found", zap.String("key", key))
+		return false
+	}
+
+	switch typeAttr := attrVal.Type(); typeAttr {
```
janatrelakova, line 96: not sure which types we want to support for resource attributes, these made sense to me, so there they are...

connector/solarwindsentityconnector/internal/log_builder_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	destination := plog.NewLogRecord()
+	setEntityType(&destination, "testEntity")
+
+	attrs, exists := destination.Attributes().Get(swoEntityType)
+	assert.True(t, exists)
+	assert.Equal(t, "testEntity", attrs.Str())
+}
+
+func TestPutStringAttribute(t *testing.T) {
```
janatrelakova, line 80: These putAttribute test could be probably improved.

connector/solarwindsentityconnector/connector.go
```diff
@@ -24,17 +27,34 @@ import (
...
 
 func (s *solarwindsentity) ConsumeMetrics(ctx context.Context, metrics pmetric.Metrics) error {
 	logs := plog.NewLogs()
+	events := internal.NewEvents(logs)
+	for i := range metrics.ResourceMetrics().Len() {
+		resourceMetric := metrics.ResourceMetrics().At(i)
+		resourceAttrs := resourceMetric.Resource().Attributes()
+
+		// This will be replaced with actual logic when conditions are introduced
+		events.AppendEntityUpdateEvent(s.entities.GetEntity("Snowflake"), resourceAttrs)
```
janatrelakova, line 51: `resourceAttrs` are used in entity schemas for inferring, so it should be enough for us to look for id attributes in resource attributes only.

connector/solarwindsentityconnector/internal/log_builder_test.go
akumetsu183: I tried parametrizing it, but it's ugly. I prefer this even if it not completely DRY.

connector/solarwindsentityconnector/generated_component_test.go
```diff
@@ -61,19 +62,19 @@ func TestComponentLifecycle(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name+"-shutdown", func(t *testing.T) {
-			c, err := tt.createFn(context.Background(), connectortest.NewNopSettings(), cfg)
+			c, err := tt.createFn(context.Background(), connectortest.NewNopSettings(metadata.Type), cfg)
```
akumetsu183, line 65: Changes due to update from 119 => 123

connector/solarwindsentityconnector/internal/event.go
```diff
@@ -0,0 +1,54 @@
...
+	attributes.PutStr(swoEntityType, entity.Type)
+	// entity attributes
+	logIds := attributes.PutEmptyMap(swoEntityAttributes)
+	for _, id := range entity.Attributes {
+		copyAttribute(&logIds, id, &resourceAttrs)
+	}
+	// timestamp
+	lr.SetObservedTimestamp(pcommon.NewTimestampFromTime(time.Now()))
+
+	eventLog := buildEventLog(&logs).AppendEmpty()
```
adschr, line 52: if I am not mistaken, this will create **new** resource log with single scope **each time** per call of the update event. This resource log, however, is general and can hold multiple log records representing various entities. Thus I'm not sure if on ingestion side it would not be more efficient to create log with the scope one **once** and only add _records_ representing individual entities to be updated (which is what I believe was happening in the original implementation)? We can discuss offline.

connector/solarwindsentityconnector/internal/event.go
```diff
@@ -0,0 +1,54 @@
...
+	// event type
+	attributes.PutStr(swoEntityEventType, entityUpdateEventType)
+	// entity type
+	attributes.PutStr(swoEntityType, entity.Type)
+	// entity attributes
+	logIds := attributes.PutEmptyMap(swoEntityAttributes)
+	for _, id := range entity.Attributes {
+		copyAttribute(&logIds, id, &resourceAttrs)
+	}
+	// timestamp
```
janatrelakova, line 49: I don't like having it here... These attributes will be set like four times in the code-base. I would say it is very easily to confuse parameters names -> `otel.entity.event_as_log` vs. `otel.entity.event.type`.

Additionally, IMHO this should be the responsibility of `log_constructor` as this is the lowest level where log is configured to be sent to SWO.

janatrelakova: Additionally, it is not consistent when setting ID attributes vs setting entity attributes.
For:
- setting ID attributes we have `setIdAttributes`
- setting entity attributes we have for loop

adschr (approved):

connector/solarwindsentityconnector/connector.go
```diff
@@ -24,29 +27,53 @@ import (
...
 }
 
 func (s *solarwindsentity) ConsumeMetrics(ctx context.Context, metrics pmetric.Metrics) error {
 	logs := plog.NewLogs()
-	err := s.logsConsumer.ConsumeLogs(ctx, logs)
-	if err != nil {
-		return err
+	events := internal.BuildEventLog(&logs)
+
+	for i := range metrics.ResourceMetrics().Len() {
```
gantrior, line 47: this looks incorrect, I guess it should be like this
```suggestion
	for i := 0; i < metrics.ResourceMetrics().Len(); i++
```

connector/solarwindsentityconnector/connector.go
```diff
@@ -24,29 +27,53 @@ import (
...
 }
 
 func (s *solarwindsentity) ConsumeLogs(ctx context.Context, logs plog.Logs) error {
 	newLogs := plog.NewLogs()
-	err := s.logsConsumer.ConsumeLogs(ctx, newLogs)
-	if err != nil {
-		return err
+	events := internal.BuildEventLog(&newLogs)
+
+	for i := range logs.ResourceLogs().Len() {
```
gantrior, line 66: the same as above

connector/solarwindsentityconnector/connector.go
janatrelakova: It is strange, it should be incorrect but it works. However, I have fixed it to the standard form.

adschr (approved):

PR #131, opened 2025-04-22
Add swoK8sWorkloadTypeProcessor

pstranak-sw:
#### Description
Provide description of the changes made.

#### Testing
Describe what testing was performed and which tests were added.


processor/swok8sworkloadtypeprocessor/processor.go
```diff
@@ -0,0 +1,224 @@
...
+			for mIdx := range metrics.Len() {
+				m := metrics.At(mIdx)
+				switch m.Type() {
+				case pmetric.MetricTypeGauge:
+					datapoints := m.Gauge().DataPoints()
+					for dpIdx := range datapoints.Len() {
+						dp := datapoints.At(dpIdx)
+						err := cp.setWorkloadTypeForGivenDatapointAttributes(dp.Attributes())
+						if err != nil {
+							return pmetric.Metrics{}, fmt.Errorf("error getting workload type for metric '%s': %w", m.Name(), err)
```
gantrior, line 103: if an error occurs when processing a metric datapoint, the entire processing stops, and all subsequent metrics/data points are skipped. Shouldn't we just log error and continue processing to avoid loosing data?

processor/swok8sworkloadtypeprocessor/processor.go
```diff
@@ -0,0 +1,224 @@
...
+					}
+				case pmetric.MetricTypeHistogram:
+					datapoints := m.Histogram().DataPoints()
+					for dpIdx := range datapoints.Len() {
+						dp := datapoints.At(dpIdx)
+						err := cp.setWorkloadTypeForGivenDatapointAttributes(dp.Attributes())
+						if err != nil {
+							return pmetric.Metrics{}, fmt.Errorf("error getting workload type for metric '%s': %w", m.Name(), err)
+						}
+					}
```
gantrior, line 123: This block is pretty much duplicated through all metric types, can you extract that to functions to avoid this duplication?

processor/swok8sworkloadtypeprocessor/processor.go
```diff
@@ -0,0 +1,224 @@
...
+	client, err := cp.config.getK8sClient()
+	if err != nil {
+		return err
+	}
+	cp.k8sClient = client
+
+	cp.factory = informers.NewSharedInformerFactory(cp.k8sClient, cp.config.WatchSyncPeriod)
+
+	registerWorkloadInformer := func(workloadType K8sWorkloadType, kind string, informer cache.SharedIndexInformer) {
+		cp.informers[workloadType] = informer
```
gantrior, line 165: It is common practices that startup of these kind of instances are logged, so I would add here something like `cp.logger.Info("Informer started", zap.String("workloadType", string(workloadType)))` so that we see in collector log how many informers are actually running.

gantrior (commented):
Adding some comments. Also I would add unittests to metric processing logic, including edge cases (missing attributes, informer errors)

processor/swok8sworkloadtypeprocessor/processor.go
pstranak-sw: Yeah, I checked the contribution guidelines for the OTEL collector and they suggest to not return errors unless really necessary. So I changed the logic in this processor accordingly.

processor/swok8sworkloadtypeprocessor/processor.go
pstranak-sw: After upgrading to OTEL collector 0.123.0 and introducing some generics here, the code should be much shorter now.

processor/swok8sworkloadtypeprocessor/processor.go
pstranak-sw: I had the logging already implemented locally, but I was waiting with it (and unit tests) for an initial review of the processor's concept first.

processor/swok8sworkloadtypeprocessor/processor.go
```diff
@@ -0,0 +1,182 @@
...
+			if namespace != "" {
+				workloadKey = fmt.Sprintf("%s/%s", namespace, name)
+			} else {
+				workloadKey = name
+			}
+
+			for _, workloadType := range workloadMapping.ExpectedTypes {
+				workload, exists, err := cp.informers[workloadType].GetStore().GetByKey(workloadKey)
+				if err != nil {
+					cp.logger.Error("error getting workload from cache", zap.String("workloadType", workloadType), zap.String("workloadKey", workloadKey), zap.Error(err))
```
gantrior, line 74: nit: This situation can happen - when entity get deleted and cacheStore gets updated before telemetry about communication gets processed, so I am not sure if I would log it as error, warn would be probably more appropriate.

cmd/solarwinds-otel-collector/components_k8s.go
```diff
@@ -116,6 +117,7 @@ func components() (otelcol.Factories, error) {
 		memorylimiterprocessor.NewFactory(),
 		batchprocessor.NewFactory(),
 		k8seventgenerationprocessor.NewFactory(),
+		swok8sworkloadtypeprocessor.NewFactory(),
```
pstranak-sw, line 120: This processor is currently registered only in the `k8s` distro image. Do we want to include it in the main distro, too? Or should we do it later?

gantrior: Yes, main distro should have all the capabilities, so you should register it there too

gantrior (approved):

cmd/solarwinds-otel-collector/components_k8s.go
pstranak-sw: changed

adschr (dismissed):

gantrior (dismissed):

PR #143, opened 2025-05-02
feature: #127 Send updates for relationships

janatrelakova:
#### Description
Support relationships for different types of entities. 
- There is some refactoring
  - `constants.go` containing attributes names needed for SWO ingestion, and the swo prefix is removed because it felt that it is polluting the code
  - `models.go` contains entity and relationship types
  - The way of setting attributes changed a little, it should be better reusable when `AppendDeleteEntityEvent` and `AppendDeleteRelationshipEvent` where we could use `CreateEntityEvent` and just set the attribute `event.type` to delete action or example
- Relationships work in the same manner as entities with different attributes for ingestion
- For relationship to work, the entities have to be defined in the schema.
- I have added for loops for entities and relationships
  - Kubernetes team wants to add a filter processor before this receiver where they pre-filter metrics and logs that will ingest chosen entities and relationships so it would not create that big load on the connector without conditions

**Tracking Issue:** https://github.com/solarwinds/solarwinds-otel-collector-releases/issues/127

Example config:
```
connectors:
  solarwindsentity:
    schema:
      entities:
        - entity: KubernetesNamespace
          id:
            - k8s.namespace.name
            - sw.k8s.cluster.uid
          attributes:
        - entity: KubernetesCluster
          attributes:
            - k8s.cluster.name
          id:
            - sw.k8s.cluster.uid
      relationships:
        - type: MemberOf
          source_entity: KubernetesCluster
          destination_entity: KubernetesNamespace
          attributes:
```

#### Testing
Add tests.

#### TODO
- [x] Add testing for relationships
- [x] Review
- [x] you can test it, but verify :-)
- [x] I didn't test optional relationship attributes, that have to be covered


connector/solarwindsentityconnector/schema.go
```diff
@@ -26,8 +27,13 @@ func (s *Schema) NewEntities() map[string]internal.Entity {
...
 			IDs:        entity.IDs,
-			Attributes: entity.Attributes}
+			Attributes: entity.Attributes,
+		}
 	}
 
 	return entities
 }
+
+func (s *Schema) NewRelationships() []internal.Relationship {
```
jirkajanecek, line 37: Mater of taste (changes up to your liking): name of `NewRelationships` inclines to creating something new. 
- In this particular case nothing new is created only returning public `Schema` property. 
- No mocking is used to confirm this approach.

connector/solarwindsentityconnector/schema.go
```diff
@@ -26,8 +27,13 @@ func (s *Schema) NewEntities() map[string]internal.Entity {
 		entities[entity.Type] = internal.Entity{
```
jirkajanecek, line 27: Matter of taste: Would you mind to pass existing instance of entity to its dedicated place in map instead of creating its copy?

connector/solarwindsentityconnector/internal/models.go
```diff
@@ -0,0 +1,14 @@
+package internal
```
jirkajanecek, line 1: Proposal (matter of taste):
- creating of `config` package containing all those models in dedicated files would increase readability.
    - at least reader will find config objects in one place (aka `Entity`, `Relationship`, `Schema`)
- it would also eliminate usage of `internal.Entity` in public code code like configuration.

connector/solarwindsentityconnector/schema.go
```diff
@@ -17,7 +17,8 @@ package solarwindsentityconnector
 import "github.com/solarwinds/solarwinds-otel-collector-releases/connector/solarwindsentityconnector/internal"
 
 type Schema struct {
```
jirkajanecek, line 19: I can see slight diversion from ADR schema [concept](https://swicloud.atlassian.net/wiki/spaces/POBP/pages/5338956937/ADR-0048+New+OTEL+Component+handling+entities+relationships+in+unified+OTEL+Collector#Usages%3A) described:
```yaml
schema:
  entities:
  events:
    relationships:
```
instead of our current implementation:
```yaml
schema:
  entities:
  relationships:
```
If it is alright, then we will face probably duplicating config issue in ADR section `schema >> events >> entities`

connector/solarwindsentityconnector/schema.go
janatrelakova: Yes, you are correct. I went with this approach to have it unified, which imho is better for the reader. 

It is not that important, we can remove the NewRelationships method.

connector/solarwindsentityconnector/connector.go
```diff
@@ -66,9 +70,12 @@ func (s *solarwindsentity) ConsumeLogs(ctx context.Context, logs plog.Logs) erro
 	for i := 0; i < logs.ResourceLogs().Len(); i++ {
 		resourceLog := logs.ResourceLogs().At(i)
 		resourceAttrs := resourceLog.Resource().Attributes()
+		internal.AppendEntityUpdateEvent(events, s.entities["Snowflake"], resourceAttrs)
```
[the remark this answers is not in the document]
janatrelakova: I would leave it like this for now. I am not sure what consequences will introduction of OTTL parsing bring.

connector/solarwindsentityconnector/connector.go
```diff
@@ -50,6 +51,9 @@ func (s *solarwindsentity) ConsumeMetrics(ctx context.Context, metrics pmetric.M
 
 		// This will be replaced with actual logic when conditions are introduced
 		internal.AppendEntityUpdateEvent(events, s.entities["Snowflake"], resourceAttrs)
+		for r := 0; r < len(s.relationships); i++ {
```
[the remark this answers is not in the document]
janatrelakova: Discussion in the previous PR https://github.com/solarwinds/solarwinds-otel-collector-releases/pull/128#discussion_r2066955606.

I will check the letters `i` and `r`. Thanks for the catch.

connector/solarwindsentityconnector/schema.go
janatrelakova: Fixed

connector/solarwindsentityconnector/internal/models.go
janatrelakova: I have created `config` package.
- I am not entirely sure, if `config` is the best name, other components do not have `config.go` and `config` directory on one level. But it is no big deal to rename it later if there is demand for that.
- I did not create file per type, I would leave it to next iteration, when we see the whole schema and all types, so it won't create overhead now.
- I have moved schema to config, because if we have all types used by schema there and config uses schema it makes sense to me to have it in one place.

connector/solarwindsentityconnector/schema.go
janatrelakova: Fixed

connector/solarwindsentityconnector/connector.go
```diff
@@ -44,12 +46,19 @@ func (s *solarwindsentity) ConsumeMetrics(ctx context.Context, metrics pmetric.M
...
 	events := internal.BuildEventLog(&logs)
 
-	for i := 0; i < logs.ResourceLogs().Len(); i++ {
+	for i := 0; i < metrics.ResourceMetrics().Len(); i++ {
 		resourceMetric := metrics.ResourceMetrics().At(i)
 		resourceAttrs := resourceMetric.Resource().Attributes()
 
 		// This will be replaced with actual logic when conditions are introduced
-		internal.AppendEntityUpdateEvent(events, s.entities["Snowflake"], resourceAttrs)
+		for k := range s.entities {
```
jirkajanecek, line 54:
```suggestion
		for _, entity := range s.entities {
```

connector/solarwindsentityconnector/connector.go
```diff
@@ -44,12 +46,19 @@ func (s *solarwindsentity) ConsumeMetrics(ctx context.Context, metrics pmetric.M
...
+			internal.AppendEntityUpdateEvent(events, s.entities[k], resourceAttrs)
```
jirkajanecek, line 55:
```suggestion
			internal.AppendEntityUpdateEvent(events, entity, resourceAttrs)
```

jirkajanecek: If there is no reason against it, I would suggest such changes for relationships as well.

jirkajanecek (dismissed):
Except small suggestions it looks good.

jirkajanecek (approved):

gantrior (approved):

Issue #144, opened 2025-05-05
Separate connection-check to a separate binary


akumetsu183:
### Is your feature request related to a problem? Please describe.

Separate connection-check to a separate binary

Create new binary “otel-endpoint-connection-check”,

it should be part of the k8s images, so they can use it in [here](https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/633c50e58bc01590ff9194b6a95978393d44a13d/deploy/helm/templates/metrics-deployment.yaml#L87)

remove old connection check code from collector binary

### Describe the solution you'd like

-

PR #145, opened 2025-05-05
feature: #144 separate binary for connection-check

akumetsu183:
#### Description
Separating connection-check to its own binary. We are doing this so that we can use the vanilla builder later on.
Binary is published as part of the k8s docker images. It will be used for checking otel endpoint connection before starting up the collector containers.

#### Testing
Describe what testing was performed and which tests were added.

### DIFF
--- a/cmd/solarwinds-otel-collector/main.go
+++ b/cmd/solarwinds-otel-collector/main.go
@@ -63,7 +63,7 @@ func main() {
 				},
 			},
 		},
-	}, logger)
+	})
 
 	if err != nil {
 		logger.Fatal("collector server run finished with error", zap.Error(err))
--- a/cmd/solarwinds-otel-collector/main_others.go
+++ b/cmd/solarwinds-otel-collector/main_others.go
@@ -18,9 +18,8 @@ package main
 
 import (
 	"go.opentelemetry.io/collector/otelcol"
-	"go.uber.org/zap"
 )
 
-func run(params otelcol.CollectorSettings, logger *zap.Logger) error {
+func run(params otelcol.CollectorSettings) error {
 	return runInteractive(params)
 }
--- a/cmd/solarwinds-otel-collector/main_windows.go
+++ b/cmd/solarwinds-otel-collector/main_windows.go
@@ -21,12 +21,11 @@ import (
 	"fmt"
 
 	"go.opentelemetry.io/collector/otelcol"
-	"go.uber.org/zap"
 	"golang.org/x/sys/windows"
 	"golang.org/x/sys/windows/svc"
 )
 
-func run(params otelcol.CollectorSettings, logger *zap.Logger) error {
+func run(params otelcol.CollectorSettings) error {
 	// No need to supply service name when startup is invoked through
 	// the Service Control Manager directly.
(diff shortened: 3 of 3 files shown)


build/docker/Dockerfile.Windows-2022
```diff
@@ -9,10 +9,12 @@ ARG CGO_ENABLED=0
 ARG GOEXPERIMENT=boringcrypto
 
 RUN cd /src/cmd/solarwinds-otel-collector && go build -tags full -trimpath -o /src/bin/solarwinds-otel-collector "-ldflags=-s -w"
+RUN cd /src/cmd/connection-check && go build -trimpath -o /src/bin/connection-check "-ldflags=-s -w"
```
jirkajanecek, line 12: I believe, we do not need connection check in non k8s docker, right? I understand connection check as proprietary addon to k8s distro.

cmd/connection-check/go.mod
```diff
@@ -0,0 +1,36 @@
+module github.com/solarwinds/solarwinds-otel-collector-releases/connection-check
+
+go 1.24.2
+
+require (
+	github.com/solarwinds/solarwinds-otel-collector-releases/pkg/version v0.123.1
+	github.com/spf13/cobra v1.9.1
```
jirkajanecek, line 7: Is it possible this package is not needed?

cmd/connection-check/main.go
```diff
@@ -86,25 +86,18 @@ func (d *OtelErrorHandler) Handle(err error) {
...
-	testCommand.MarkFlagRequired("clusteruid")
-	testCommand.MarkFlagRequired("endpoint")
-	testCommand.MarkFlagRequired("apitoken")
-
-	return testCommand
+	defer logger.Sync()
+
+	clusterUid := flag.String("clusteruid", "", "Cluster UID")
+	endpoint := flag.String("endpoint", "", "Endpoint")
+	apiToken := flag.String("apitoken", "", "Api token")
```
jirkajanecek, line 98: Milane, would you consider to use `*Var` calls instead of `String()` calls?
- https://pkg.go.dev/flag#FlagSet.StringVar

There is an advantage in later utilization (line 102) to avoid referencing (imo looks strange).

build/docker/Dockerfile.Windows-2022
akumetsu183: True. My brain autofilled k8s

cmd/connection-check/go.mod
akumetsu183: go mod tidied it away

cmd/connection-check/main.go
akumetsu183: Done.

cmd/solarwinds-otel-collector/main.go
```diff
@@ -63,16 +63,15 @@ func main() {
 				},
 			},
 		},
-	}, logger)
```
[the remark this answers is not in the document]
akumetsu183: It's back. But I still don't see why it's needed.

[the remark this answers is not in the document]
akumetsu183: Aaand, it's back after further discussions.

jirkajanecek (approved):

PR #149, opened 2025-05-07
feature: #150 playground and verified distributions

akumetsu183:
#### Description
playground-and-verified distributions

Missing receivers.

-  utf8consistency
- hostinfo processor
- hostvolumes processor
- uuid processor
- uams processor


cmd/solarwinds-otel-collector/components_verified.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/oracledbreceiver"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/rabbitmqreceiver"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/redisreceiver"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/snowflakereceiver"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/statsdreceiver"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/zookeeperreceiver"
+	"go.opentelemetry.io/collector/receiver/nopreceiver"
+	// all necessary vanilla processors are already part of the k8s bunch
+	// missing utf8consistency
+	// missing hostinfo processor
```
jirkajanecek, line 96: Is it correct to leave `hostvolumes` processor behind? The processor itself adds some missing functionality for Windows and its availability would be beneficial.

I understand there might be technological reasons (processor is currently part of uams plugin), but in that case do we have some follow up task for "publishing"?

cmd/solarwinds-otel-collector/components_verified.go
```diff
@@ -0,0 +1,198 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build verified
+
+package main
+
+import (
+	// all ours
```
jirkajanecek, line 20: I would consider to hold the same component structures as it in `playground`.  Grouped by component types (receivers, processors, generals, ...).

I also think comments describing reasoning behind components un/availability should be place into this PR not the code.

.github/workflows/buildAndTest.yml
```diff
@@ -42,32 +42,59 @@ jobs:
...
         uses: ./.github/actions/dockerScoutScan
         with:
           username: ${{ vars.ENOPS5919_DOCKER_SCOUT_CI_USER }}
           token: ${{ secrets.ENOPS5919_DOCKER_SCOUT_CI_PAT }}
           current-image: solarwinds-otel-collector
-          current-tag: ${{ needs.generate_tag.outputs.tag }}
+          current-tag: ${{ needs.generate_tag.outputs.tag }}-playground
           compare-to-image: ${{ env.DOCKERHUB_IMAGE }}
-          compare-to-tag: latest
+          compare-to-tag: latest # to be updated
```
jirkajanecek, line 70: Updated when? After the first successful release with this new tags?

.github/workflows/deploy.yml
```diff
@@ -85,7 +85,20 @@ jobs:
...
+        with:
+          context: .
+          file: build/docker/Dockerfile
+          platforms: linux/amd64,linux/arm64
+          provenance: false
+          sbom: false
+          push: true
+          tags: |
+            ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}
+            ${{ env.DOCKERHUB_IMAGE }}:latest-verified
```
jirkajanecek, line 101: Just a question to think off. Now I understand we have two tags for latest:
- `:latest-playground`
- `:latest-verified`

this looks to me like new convention we have. Shall we consider to add also `:latest-k8s`?

jirkajanecek (dismissed):
It looks fine to me. Questions and ideas are merely cosmetic for more unified and complete feeling.

.github/workflows/buildAndTest.yml
akumetsu183: After first publish,... successful publish. We need the tags to be in dockerhub.

.github/workflows/deploy.yml
adschr: Sounds good to me. Nevetheless, I'd also mark the 'verified' as 'latest' only, as it should be the 'main' distribution?
We'll also need to update the manifest mojo (See ie. the Get Manifest for ... image) - update it not to contain full, but `verified` and `playground`

jirkajanecek: Regarding `latest-k8s`, Milan has told me, k8s team is not eager to have `latest` tag on their images. Reasoning behind this is a need to have "always versioned" images deployed -?? most likely for better auditing and control in cluster??...

Regarding `latest` tag  for `verified` as main distribution is not probably against nothing. But let's be more descriptive, does it mean following tags?:
- `verfied`, `latest`
- `playground`

akumetsu183: Updated, added :latest (to verified). Added latest-k8s

jirkajanecek (dismissed):

jirkajanecek (approved):

Issue #150, opened 2025-05-07
Create new distributions


akumetsu183:
### Is your feature request related to a problem? Please describe.

Aim of this task is to:

create new distribution with the current components set

called “-playground“

update full distribution to contain only the components required for SWO integration

called “-verified“

### Describe the solution you'd like

---

PR #152, opened 2025-05-13
feature: #150 align latest tags

akumetsu183:
#### Description
feature: #150 align latest tags

Tested
image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -146,12 +146,12 @@ jobs:
 
       - name: Tag images
         run: |
-          docker tag solarwinds-otel-collector:${{ steps.generate-tag.outputs.tag }}-nanoserver-ltsc2022-playground ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2022-playground
-          docker tag solarwinds-otel-collector:${{ steps.generate-tag.outputs.tag }}-nanoserver-ltsc2019-playground ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2019-playground
-          docker tag solarwinds-otel-collector:${{ steps.generate-tag.outputs.tag }}-nanoserver-ltsc2022-verified ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2022-verified
-          docker tag solarwinds-otel-collector:${{ steps.generate-tag.outputs.tag }}-nanoserver-ltsc2019-verified ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2019-verified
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -197,12 +197,12 @@ jobs:
...
-          docker push ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2019-verified
-          docker push ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2022-k8s
-          docker push ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2019-k8s
+          docker push --all-tags ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2022-playground
+          docker push --all-tags ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2019-playground
+          docker push --all-tags ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2022-verified
+          docker push --all-tags ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2019-verified
+          docker push --all-tags ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2022-k8s
+          docker push --all-tags ${{ env.DOCKERHUB_IMAGE }}:${{ needs.release_checks.outputs.image_tag }}-nanoserver-ltsc2019-k8s
```
akumetsu183, line 205: 🤷 
https://docs.docker.com/reference/cli/docker/image/push/#all-tags

adschr: 🤔 if I understand it correctly:
- **Tag images** actually does two kinds of tagging: 
  - `solarwinds-otel-collector:*` - adds multiple tags (image:sometag:sometag) - with version and latest
  - `${{ env.DOCKERHUB_IMAGE}}:*` - still adds single tag - just with version
- **Push as specific**, we enforce --all-tags for the `${{ env.DOCKERHUB_IMAGE}}`.

So if we want to publish latest an version to dockerhub I think we need to:
- update the **Tag images** to add multiple tags to `${{ env.DOCKERHUB_IMAGE}}:*` instead
- update the **Push as specific** to docker push --all-tags ${{ env.DOCKERHUB_IMAGE }}; not sure if it is verified to work as used here (--all-tags, but use image + some tag for the arg? docs seem to mention only image as the arg)

But feel free to correct me if I am wrong in my hypotheses!

adschr (commented):
Naming unification LGTM! Left a question on the win images publishing.

.github/workflows/deploy.yml
adschr: I'd also check if we use the `solarwinds-otel-collector:` variants anywhere else. If not, I'd drop those to reduce complexity and thus potential confusion. It was originally used on testing for linux images which does not happen for windows AFAIK; but maybe it is utilized by the analysis actions.

akumetsu183: I... re-worked it, to make it bit more simple and hopefully, correct.

adschr (approved):

PR #156, opened 2025-05-15
Feature/send update same type relationship

KulinkovichAndrei-SE:
#### Description
Added support for creating events for same-type relationships.

This includes extending the configuration parsing with _source_ and _destination_ ID attribute prefixes to define the direction of same-type relationships.

A new _EventBuilder_ structure was added to encapsulate entity, relationship schema and prefixes value creation based on the config.

The event creation logic was updated to separate the processing of default and same-type relationships. This led to the splitting of the _setIdAttributes_ function into two: one for default cases and one for same-type relationships.

For same-type relationships, a relationship is created only if at least one ID attribute key on both entities has the configured prefix. This allows support for unprefixed common ID attributes and avoid ambiguous matches when direction is unclear.

#### Testing
Unit and integration tests were updated and extended to cover the new functionality. The functionality was tested using new unit and integration tests, as well as by verifying data in the DEV SWO environment and checking logs via Loggly.


janatrelakova:
Please, could you fix the failing build?

Additionally, could we add some more information to the description?
- how the configuration looks like (from where we are obtaining the prefixes) so it is more readable for the reviewer?
- examples of generated same-type relationships or the resource attributes needed -> why 1xsource without dest does not work and why 2xsource + 1x dest works, so it is clear what scenarios we support
- could you add here or in the JIRA how the reviewer is able to test it, where is he supposed to look for ingested relationship?

See older PR for the reference.

connector/solarwindsentityconnector/connector.go
```diff
@@ -43,47 +49,47 @@ func (s *solarwindsentity) Capabilities() consumer.Capabilities {
...
-	events := internal.BuildEventLog(&logs)
+	newLogs := plog.NewLogs()
+	eventBuilder := internal.NewEventBuilder(s.entities, s.relationships, s.sourcePrefix, s.destinationPrefix, &newLogs, s.logger)
 
 	for i := 0; i < metrics.ResourceMetrics().Len(); i++ {
 		resourceMetric := metrics.ResourceMetrics().At(i)
 		resourceAttrs := resourceMetric.Resource().Attributes()
 
 		// This will be replaced with actual logic when conditions are introduced
 		for _, entity := range s.entities {
```
adschr, line 60: we iterate over entities and relationships, but event builder has access to that static data. Did you consider moving the loop into the event builder, passing only the resource attrs as an input?

janatrelakova: Not sure, if it is worth it (maybe it is), I would wait for the conditions.

connector/solarwindsentityconnector/connector_test.go
```diff
@@ -45,26 +46,63 @@ func (t *testLogsConsumer) ConsumeLogs(ctx context.Context, ld plog.Logs) error
 
 var _ consumer.Logs = (*testLogsConsumer)(nil)
 
+// Test configuration for the entites, relationships and events.
 var (
-	expectedEntities = []config.Entity{
-		{Type: "Snowflake", IDs: []string{"id1"}, Attributes: []string{"attr1"}},
+	configuredEntities = []config.Entity{
+		{Type: "Snowflake", IDs: []string{"snowflake.id"}, Attributes: []string{"attr1"}},
+		{Type: "AWS", IDs: []string{"aws.ec2.id", "aws.ec2.name"}, Attributes: []string{"attr2"}},
```
janatrelakova, line 53: nit
```suggestion
		{Type: "AWS EC2", IDs: []string{"aws.ec2.id", "aws.ec2.name"}, Attributes: []string{"attr2"}},
```

connector/solarwindsentityconnector/connector.go
```diff
@@ -43,47 +49,47 @@ func (s *solarwindsentity) Capabilities() consumer.Capabilities {
 }
 
 func (s *solarwindsentity) ConsumeMetrics(ctx context.Context, metrics pmetric.Metrics) error {
-	logs := plog.NewLogs()
-	events := internal.BuildEventLog(&logs)
+	newLogs := plog.NewLogs()
+	eventBuilder := internal.NewEventBuilder(s.entities, s.relationships, s.sourcePrefix, s.destinationPrefix, &newLogs, s.logger)
```
adschr, line 53: did you consider having the event builder as singleton defined within the connector struct? I understand that we construct it each time only because of the different `&newLogs`. My line of thought is that I'd either:
- have single eventBuilder for the connector, declared with static structures (entities, relationship, logger, ...) => pass the logs ref via the function calls
- perceive the event builder as event builder for single stream of events - thus it would not need &newLogs input, it would be constructed within and obtained via some accessor when required (for lns 70+)

connector/solarwindsentityconnector/internal/event.go
```diff
@@ -15,47 +15,135 @@
...
+	lr := plog.NewLogRecord()
+	attrs := lr.Attributes()
 
-	eventLog := lrs.AppendEmpty()
-	event.CopyTo(eventLog)
+	if source.Type == dest.Type {
+		// same type relationships
+		hasPrefixSrc, err := setIdAttributesSameType(attrs, source.IDs, resourceAttrs, relationshipSrcEntityIds, e.SourcePrefix)
+		if err != nil {
+			return plog.NewLogRecord(), fmt.Errorf("bad source entity id attributes")
```
janatrelakova, line 120:
```suggestion
			return plog.NewLogRecord(), fmt.Errorf("missing ID attribute for source entity")
```

connector/solarwindsentityconnector/internal/log_builder.go
```diff
@@ -16,44 +16,62 @@ package internal
...
-}
-
-// setIdAttributes sets the entity id attributes in the log record as needed by SWO.
+// setIdAttributesDefault sets the entity id attributes in the log record as needed by SWO.
 // Attributes are used to infer the entity in the system.
 //
 // Returns error if any of the attributes are missing in the resourceAttrs.
 // If any ID attribute is missing, the entity would not be inferred.
-func setIdAttributes(attrs pcommon.Map, entityIds []string, resourceAttrs pcommon.Map, name string) error {
+func setIdAttributesDefault(attrs pcommon.Map, entityIds []string, resourceAttrs pcommon.Map, name string) error {
```
janatrelakova, line 28: What Default means? could we name this methods better?

connector/solarwindsentityconnector/connector.go
adschr: ok, if we work with the premise that sth out of scope of the event builder will happen (evaluation of whether the event builder will be even invoked) let's keep as is.

connector/solarwindsentityconnector/internal/event.go
```diff
@@ -15,47 +15,135 @@
...
+		Relationships:  relationships,
+		SourcePrefix:   sourcePrefix,
+		DestPrefix:     destPrefix,
+		ResultEventLog: CreateResultEventLog(events),
+		Logger:         logger,
+	}
+}
+
+// CreateResultEventLog prepares a clean LogRecordSlice, where log records representing events should be appended.
+// In given plog.Logs creates a resource log with one scope log and set attributes needed by SWO ingestion.
```
adschr, line 48:
```suggestion
// Creates a resource log in input plog.Logs with single scope log decorated with attributes necessary for proper SWO ingestion.
```

connector/solarwindsentityconnector/testdata/logsToLogs/relationship/expected-log-different-type-relationship.yaml
```diff
@@ -0,0 +1,88 @@
...
+              - key: otel.entity.type
+                value:
+                  stringValue: Snowflake
+              - key: otel.entity.id
+                value:
+                  kvlistValue:
+                    values:
+                      - key: snowflake.id
+                        value:
+                          stringValue: snowflake-instance-id-first
```
janatrelakova, line 15: Huge nit: 
```suggestion
                          stringValue: snowflake-instance-id
```

for all values, so we keep these yaml files clean as possible.

connector/solarwindsentityconnector/internal/event.go
```diff
@@ -15,47 +15,135 @@
...
 )
 
-func AppendEntityUpdateEvent(lrs *plog.LogRecordSlice, entity config.Entity, resourceAttrs pcommon.Map) {
-	event, err := CreateEntityEvent(resourceAttrs, entity)
+type EventBuilder struct {
+	Entities       map[string]config.Entity
+	Relationships  []config.Relationship
+	SourcePrefix   string
+	DestPrefix     string
+	ResultEventLog *plog.LogRecordSlice
```
adschr, line 32: does this need to be public? This prop is crucial for creating the resulting events and is initiated in special manner. Having it public means it can be rewritten from the outside leading to broken result. Not that I'd expect it to be real issue in this particular case, but I'd rather limit the access and if necessary, expose special method for read access.

connector/solarwindsentityconnector/internal/event.go
```diff
@@ -15,47 +15,135 @@
...
+
+	setAttributes(attrs, entity.Attributes, resourceAttrs, entityAttributes)
+
+	lr.SetObservedTimestamp(pcommon.NewTimestampFromTime(time.Now()))
+
+	return lr, nil
+}
+
+func (e *EventBuilder) AppendRelationshipUpdateEvent(relationship config.Relationship, resourceAttrs pcommon.Map) {
+	relationshipLog, err := e.createRelationshipEvent(relationship, resourceAttrs)
```
adschr, line 89: nit: unify naming with ApendEntityUpdateEvent? `relationshipLog` vs `event`

connector/solarwindsentityconnector/internal/event.go
```diff
@@ -15,47 +15,135 @@
...
-	src, ok := entities[relationship.Source]
-	if !ok {
-		zap.L().Debug("source entity not found", zap.String("entity", relationship.Source))
+func (e *EventBuilder) createEntityEvent(resourceAttrs pcommon.Map, entity config.Entity) (plog.LogRecord, error) {
+	lr := plog.NewLogRecord()
+	attrs := lr.Attributes()
+	attrs.PutStr(entityType, entity.Type)
+
+	if err := setIdAttributesDefault(attrs, entity.IDs, resourceAttrs, entityIds); err != nil {
+		return plog.LogRecord{}, fmt.Errorf("failed to set id attributes: %w", err)
```
adschr, line 78: same for relationships - does returning empty log records have any significance? can we propagate nil + err?

connector/solarwindsentityconnector/internal/event.go
```diff
@@ -15,47 +15,135 @@
...
+		if err != nil {
+			return plog.NewLogRecord(), fmt.Errorf("bad source entity id attributes")
+		}
+
+		hasPrefixDst, err := setIdAttributesSameType(attrs, dest.IDs, resourceAttrs, relationshipDestEntityIds, e.DestPrefix)
+		if err != nil {
+			return plog.NewLogRecord(), fmt.Errorf("bad destination entity id attributes")
+		}
+
+		if !hasPrefixSrc || !hasPrefixDst {
```
adschr, line 128: Maybe I'm missing something, but since we use or anyway and work with no particular info afterwards, would it make to fail fast? maybe propagate the `!hasPrefix` state as an error instead of separate bool value?

connector/solarwindsentityconnector/internal/log_builder.go
```diff
@@ -16,44 +16,62 @@ package internal
...
+			return fmt.Errorf("missing entity ID attribute: %s", id)
 		}
+
 		putAttribute(&logIds, id, value)
 	}
 	return nil
 }
 
+// setIdAttributesSameType sets the entity id attributes in the log record as needed by SWO for same type relationships.
+// Verifies that prefix is present in the resource attributes at least once for each entities.
```
adschr, line 47:
```suggestion
// Verifies that prefix is present in the resource attributes at least once within entity IDs.
```
Is this 'at least once' just safety mechanism to make sure the relationships vs. entity ids is done correctly, or does it have some higher meaning? Maybe I'd add a little note on why this check is done.

connector/solarwindsentityconnector/internal/log_builder_test.go
```diff
@@ -258,6 +347,125 @@ func TestCreateRelationshipEventWithoutResourceAttributes(t *testing.T) {
...
+	resourceAttrs.PutStr("attr1", "attrvalue1")
+
+	entity := config.Entity{
+		Type:       "KubernetesCluster",
+		IDs:        []string{"id"},
+		Attributes: []string{"attr1"},
+	}
+
+	relationship := config.Relationship{
+		Type:        "VirtualizationTopologyConnection",
```
janatrelakova, line 380: Very minor: Could we use something like `TestRelationshipType` instead of `VirtualizationTopologyConnection`.

Motivation: this probably does not work in our system and it could be confusing for NIO team (if they ever see this) why we are using their relationship type. Having `TestRelationshipType` is perfectly fine IMHO.

connector/solarwindsentityconnector/internal/log_builder.go
janatrelakova, line 1: As discussed before, could we rename this file to better reflect what is going on. Only some attributes helpers stayed here.

connector/solarwindsentityconnector/internal/event.go
```diff
@@ -15,47 +15,135 @@
...
+		SourcePrefix:   sourcePrefix,
+		DestPrefix:     destPrefix,
+		ResultEventLog: CreateResultEventLog(events),
+		Logger:         logger,
+	}
+}
+
+// CreateResultEventLog prepares a clean LogRecordSlice, where log records representing events should be appended.
+// In given plog.Logs creates a resource log with one scope log and set attributes needed by SWO ingestion.
+func CreateResultEventLog(logs *plog.Logs) *plog.LogRecordSlice {
```
janatrelakova, line 49:
```suggestion
func createResultEventLog(logs *plog.Logs) *plog.LogRecordSlice {
```
this is used only when creating a new Event builder.

connector/solarwindsentityconnector/testdata/logsToLogs/relationship/expected-log-same-type-relationship-nomatch.yaml
janatrelakova, line 1: What do you think of this structure:
- logsToLogs
  - relationship
    -  different-types-relationship
        - input-log.yaml
        - expected-log.yaml
    - same-type-relationship
       - input-log.yaml
      - expected-log.yaml
  - entity ....
  
  
  Change it, if you like it.... The positive would be better navigation between input and expected log.

connector/solarwindsentityconnector/testdata/logsToLogs/relationship/input-log-same-type-relationship.yaml
```diff
@@ -0,0 +1,13 @@
...
+                    { "key": "aws.ec2.name", "value":{"stringValue": "aws-instance-name-first"}}]
+  - resource:
+      attributes: [ { "key": "aws.ec2.id", "value":{"stringValue": "aws-instance-id-second"}},
+                    { "key": "aws.ec2.name", "value":{"stringValue": "aws-instance-name-second"}}]                        
+  - resource:
+      attributes: [ { "key": "src.aws.ec2.id", "value":{"stringValue": "aws-instance-id-first"}},
+                    { "key": "src.aws.ec2.name", "value":{"stringValue": "aws-instance-name-first"}},
+                    { "key": "dst.aws.ec2.id", "value":{"stringValue": "aws-instance-id-second"}},
+                    { "key": "dst.aws.ec2.name", "value":{"stringValue": "aws-instance-name-second"}}]
+
```
janatrelakova, line 13: If we could add one common attribute, it would be great to test it thoroughly.

for example:
                    { "key": "aws.ec2.common.attribute", "value":{"stringValue": "common-value}}]

adschr (dismissed):
Overall flow LGTM - left just some design questions (mainly eventBuilder construction and propagation of `hasPrefix` bools) for consideration; otherwise mainly naming/docs nitpicks with minor importance

connector/solarwindsentityconnector/internal/event.go
```diff
@@ -15,47 +15,135 @@
...
+	Logger         *zap.Logger
+}
+
+func NewEventBuilder(entities map[string]config.Entity, relationships []config.Relationship, sourcePrefix string, destPrefix string, events *plog.Logs, logger *zap.Logger) *EventBuilder {
+	return &EventBuilder{
+		Entities:       entities,
+		Relationships:  relationships,
+		SourcePrefix:   sourcePrefix,
+		DestPrefix:     destPrefix,
+		ResultEventLog: CreateResultEventLog(events),
```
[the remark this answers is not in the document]
KulinkovichAndrei-SE: What exactly should be renamed? ResultEventLog?

connector/solarwindsentityconnector/internal/log_builder.go
janatrelakova: Ok, but is not that clear for the reader what default means in this context.

janatrelakova:
Testing seems fine to me, I would maybe add some scenarios:
- different types relationship where there are redundant resource attribute not used for any entity (just a check we don't fail if there is something redundant)
- different types relationship where one id attribute is missing
- relationship (same/different types) with using relationship attributes (relationship attributes are defined in the configuration)

connector/solarwindsentityconnector/internal/event.go
KulinkovichAndrei-SE: cannot use nil as plog.LogRecord value in return statement

connector/solarwindsentityconnector/internal/log_builder.go
```diff
@@ -16,44 +16,62 @@ package internal
...
+func setIdAttributesSameType(attrs pcommon.Map, entityIds []string, resourceAttrs pcommon.Map, name, prefix string) (bool, error) {
+	if len(entityIds) == 0 {
+		return false, fmt.Errorf("entity id attributes are empty")
```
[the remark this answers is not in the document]
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/internal/event.go
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/connector.go
```diff
@@ -43,47 +49,47 @@ func (s *solarwindsentity) Capabilities() consumer.Capabilities {
 }
 
 func (s *solarwindsentity) ConsumeMetrics(ctx context.Context, metrics pmetric.Metrics) error {
-	logs := plog.NewLogs()
-	events := internal.BuildEventLog(&logs)
+	newLogs := plog.NewLogs()
```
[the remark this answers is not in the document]
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/internal/event.go
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/internal/log_builder.go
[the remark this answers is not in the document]
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/internal/log_builder.go
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/internal/event.go
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/testdata/logsToLogs/relationship/expected-log-different-type-relationship.yaml
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/internal/event.go
KulinkovichAndrei-SE: fixed, now private

connector/solarwindsentityconnector/internal/log_builder.go
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/internal/log_builder_test.go
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/internal/log_builder.go
KulinkovichAndrei-SE: fixed, the file is now named attribute_helper.go and attribute_helper_test.go

connector/solarwindsentityconnector/internal/event.go
KulinkovichAndrei-SE: fixed

connector/solarwindsentityconnector/connector.go
KulinkovichAndrei-SE: As discussed during the meeting, we'll keep it as is. Marking this as resolved.

connector/solarwindsentityconnector/testdata/logsToLogs/relationship/input-log-same-type-relationship.yaml
KulinkovichAndrei-SE: added

connector/solarwindsentityconnector/testdata/logsToLogs/relationship/expected-log-same-type-relationship-nomatch.yaml
KulinkovichAndrei-SE: added same-type-relationship and different-types-relationship folders

adschr (dismissed):

adschr (approved):

PR #182, opened 2025-05-30
docs: integration templates

jirkajanecek:
#### Description
This PR adds examples of integrations configurations. Notable pointers:
- `Required parameter` marks parameter that needs to be filled in for the integration to work and start polling some data
- `Required secret` marks parameter that needs to be provided in secure way providing proper authentication to polled endpoints
- `Optional *` marks parameters and secrets that may be providedto customize final experience. These are commented-out and needs to be uncommented to be filled in (this may apply to whole sections). Some of optional sections may contain parameters becoming required after uncommenting the section (ex. filelog receiver application)

Readmes with setup steps, if needed, will be provided in the next iteration.

#### Testing
Using the configs with proper variable replacement data start arriving in correct form.


examples/integrations/apache/config.yaml
```diff
@@ -0,0 +1,157 @@
...
+    metrics:
+      exclude:
+        match_type: strict
+        metric_names:
+          - apache.request.time.rate.temp
+  memory_limiter:
+    check_interval: 1s
+    limit_percentage: 50
+    spike_limit_percentage: 30
+    detectors: ["system"]
```
adschr, line 110: If I am not mistaken, the `detectors/timeout/override/system` properties come from `resourcedetection` processor which should be used as well in all variants. Or is there plan not to include those?

jirkajanecek: That's probably happened after incorrect delete of resource detection setups. Fixed.

adschr (dismissed):

PR #206, opened 2025-07-03
chore: add supported integrations in README

david-sevcik:
#### Description
Adds a section in the README about supported integrations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ docker run  -v ./config.yaml:/opt/default-config.yaml solarwinds/solarwinds-otel
 
 ### Fully supported integrations
 
-If you'd like, you can find example configurations [here](/examples/integrations/). When these integrations are configured, you will get the same experience as with integrations configured in the managed way using Add Data wizards in the SolarWinds Observability.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,6 +63,10 @@ docker pull solarwinds/solarwinds-otel-collector:playground
 docker run  -v ./config.yaml:/opt/default-config.yaml solarwinds/solarwinds-otel-collector:playground
 ```
 
+### Fully supported integrations
+
+If you'd like, you can find example configurations [here](/examples/integrations/). When these integrations are configured, you will get the same experience as with integrations configured in the managed way using Add Data wizards in the SolarWinds Observability.
```
michal-sostak-solarwinds, line 68: just removing "the" article before Solarwinds Observability as it shouldn't be needed (checked with copilot)
```suggestion
If you'd like, you can find example configurations [here](/examples/integrations/). When these integrations are configured, you will get the same experience as with integrations configured in the managed way using Add Data wizards in SolarWinds Observability.
```

michal-sostak-solarwinds (approved):

PR #218, opened 2025-07-22
build: update deploy workflow

KulinkovichAndrei-SE:
#### Description
Removed Docker image building and pushing steps from GitHub Actions pipeline. Updated deploy workflow to verify existing Docker images instead of building new ones. Added links to each deployed image in the release notes.

#### Testing
Will be validated with the next release.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -12,21 +12,6 @@ updates:
     commit-message:
       prefix: "chore: "
 
-  - package-ecosystem: docker
-    directory: /build/docker
-    schedule:
-      interval: weekly
-      day: saturday
-    labels:
-      - dependencies
-      - docker
-    groups:
-      docker:
-        patterns:
-          - "*"
-    commit-message:
-      prefix: "chore: "
-
   - package-ecosystem: gomod
     directories:
       - "/internal/*/"


.github/workflows/deploy.yml
```diff
@@ -344,9 +102,23 @@ jobs:
             --title v${{ needs.release_checks.outputs.image_tag }} \
             --generate-notes \
             --notes-start-tag ${{ env.PREVIOUS_TAG }}
+      
+      - name: Append Docker image links to release notes
+        env:
+          # for gh cli
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release view v${{ needs.release_checks.outputs.image_tag }} --json body -q .body > notes.md
```
adschr, line 111: I presume we can't add the notes right ahead in the previous step due to `--generate-notes` being in conflict with `--notes` or `--notes-file`?

build/prepare-release.sh
```diff
@@ -34,45 +34,6 @@ else
     echo "CHANGELOG.md already contains 'v$VERSION', no update made."
 fi
 
-# Update release manifest files
-ALL_MANIFEST_YAML=$(find . -name "manifest.yaml" -type f | sort)
```
adschr, line 38: is OTEL_VERSION still used? (ln22) let's drop it also from the ln 17 reference if not.

.github/workflows/deploy.yml
KulinkovichAndrei-SE: Yes, that's correct.

adschr (commented):
Great job 👍 left two comments, before proceeding let's make sure the script input validation is corrected but besides that lookin' good!

PR #224, opened 2025-07-28
chore: delete cmd folder

KulinkovichAndrei-SE:
#### Description
Deleted cmd folder.


adschr (commented):
Please re-add the `changes-analyzer` folder as it is utilized outside this repository and removal would result in breaking change => let's remove only the `cmd/connection-check` folder

